PR #20, opened 2025-04-07
Log client errors to console.warn().

### LINKED ISSUE #19: Add logging for handled errors
e.g. all of the 401s, logging what the actual response code/error is would be valuable

we're debugging an issue right now that is getting an 'invalid client_id' response but I had to inject my own wrapper to actually figure out what the error was (it was happening in Cursor/Claude)

kentonv:
Fixes #19.

Question to consider: Is it OK to have logging on by default? I think probably yes. But it is weird that someone could create a lot of noise in the logs by sending a log of bogus requests I guess?


geelen:
Folding into #22

PR #22, opened 2025-04-09
Added onError callback

geelen:
Solves #19 in a more flexible way, by giving the user a hook to send events anywhere they wish.

We could _also_ merge #20 so these events are logged by default in a standard way? @kentonv wdyt?


kentonv (approved):
Have Claude write a test.

PR #24, opened 2025-04-24
Support multiple api handlers

geelen:
```ts
  // For multi-handler setups, you can use apiHandlers instead of apiRoute+apiHandler.
  // This allows you to use different handlers for different API routes.
  // Note: You must use either apiRoute+apiHandler (single-handler) OR apiHandlers (multi-handler), not both.
  // Example:
  // apiHandlers: {
  //   "/api/users/": UsersApiHandler,
  //   "/api/documents/": DocumentsApiHandler,
  //   "https://api.example.com/": ExternalApiHandler,
  // },
```

### DIFF
--- a/src/oauth-provider.ts
+++ b/src/oauth-provider.ts
@@ -99,15 +99,33 @@ export interface OAuthProviderOptions {
    * URL(s) for API routes. Requests with URLs starting with any of these prefixes
    * will be treated as API requests and require a valid access token.
    * Can be a single route or an array of routes. Each route can be a full URL or just a path.
+   * 
+   * Used with `apiHandler` for the single-handler configuration. This is incompatible with
+   * the `apiHandlers` property. You must use either `apiRoute` + `apiHandler` OR `apiHandlers`, not both.
    */
-  apiRoute: string | string[];
+  apiRoute?: string | string[];
 
   /**
    * Handler for API requests that have a valid access token.
    * This handler will receive the authenticated user properties in ctx.props.
    * Can be either an ExportedHandler object with a fetch method or a class extending WorkerEntrypoint.
+   * 
+   * Used with `apiRoute` for the single-handler configuration. This is incompatible with
+   * the `apiHandlers` property. You must use either `apiRoute` + `apiHandler` OR `apiHandlers`, not both.
    */
-  apiHandler: ExportedHandlerWithFetch | (new (ctx: ExecutionContext, env: any) => WorkerEntrypointWithFetch);
+  apiHandler?: ExportedHandlerWithFetch | (new (ctx: ExecutionContext, env: any) => WorkerEntrypointWithFetch);
+
+  /**
+   * Map of API routes to their corresponding handlers for the multi-handler configuration.
+   * The keys are the API routes (strings only, not arrays), and the values are the handlers.
+   * Each route can be a full URL or just a path, and each handler can be either an ExportedHandler
+   * object with a fetch method or a class extending WorkerEntrypoint.
+   * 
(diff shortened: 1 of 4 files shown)


geelen:
Published `0.0.4-0` on the `next` tag for testing

src/oauth-provider.ts
```diff
@@ -671,26 +689,68 @@ class OAuthProviderImpl {
...
    */
-  private typedApiHandler: TypedHandler;
   private typedDefaultHandler: TypedHandler;
+  
+  /**
+   * Map of API routes to their validated handlers
+   * In the simple case, this will be a single entry with the route and handler from options.apiRoute/apiHandler
+   * In the advanced case, this will contain entries from options.apiHandlers
+   */
+  private typedApiHandlers: Map<string, TypedHandler>;
```
kentonv, line 699: It looks like we never actually do a lookup in this Map, we always iterate over it. So why use a map at all? An array of tuples would be more efficient, since they wouldn't need to maintain a hash table.

src/oauth-provider.ts
```diff
@@ -896,14 +956,28 @@ class OAuthProviderImpl {
...
+  /**
+   * Finds the appropriate API handler for a URL
+   * @param url - The URL to find a handler for
+   * @returns The TypedHandler for the URL, or undefined if no handler matches
+   */
+  private findApiHandlerForUrl(url: URL): TypedHandler | undefined {
+    // Check each route in our map of validated API handlers
+    for (const [route, handler] of this.typedApiHandlers.entries()) {
+      if (this.matchApiRoute(url, route)) {
+        return handler;
```
kentonv, line 977: What if multiple routes match?

This code takes the first match -- _except_ when there are two identical routes, in which case the last match wins, due to having overwritten the earlier ones in the Map. I think at the very least we need to make this consistent. (Could be that we throw an error on duplicates.)

A "most specific route wins" policy might be more intuitive than "first match wins". We could achieve that by using an array and sorting by length (descending) upfront.

kentonv (commented):
So did AI write this or did you actually do it the old fashion way?

(If any part was AI, please say so in the commit description.)

src/oauth-provider.ts
geelen: Oh you mean duplicate keys in the literal map?

```ts
  apiHandlers: {
    '/sse': MyMCP.serveSSE('/sse'),
    '/sse': MyMCP.serve('/mcp'),
  },
```

That doesn't feel like a case the library needs to worry about, does it? That's a collision before `apiHandlers` is even received by the library—once `options.apiHandlers` is being processed there can't be any duplicates. And it's a TS error and will be flagged by any IDE or linter as being a mistake.

First-match wins is pretty standard for routing tables, too. E.g.

```
const app = new Hono()
app.get('/foo/*', () => { ... })
app.get('/foo/bar/baz', () => { ... }) // unreachable
```

Imo any kind of heuristic to reorder routes (like longest-first) is more confusing than just honouring the order the items are written in.

src/oauth-provider.ts
geelen: Fixed.

geelen:
> So did AI write this or did you actually do it the old fashion way?
> 
> (If any part was AI, please say so in the commit description.)

Updated commit description. Published under `next` dist tag, `0.0.4-1` if you want to try it out.

kentonv (approved):

src/oauth-provider.ts
kentonv: Ah yeah I forgot the input was an object so can't have duplicate keys anyway.

Amazingly, JavaScript itself apparently lets you type objects with duplicate keys. The last one wins, apparently. WTF? But that's not our problem.

PR #26, opened 2025-04-30
Validating redirect_uri on authorize

geelen:
This ends up calling `this.lookupClient(clientId)` inside `parseAuthRequest`, which I think is ok.

Annoyingly, I also usually call that from outside, but then I don't really want to change the return value of `oauthReqInfo` because I've been serialising this directly in the `state` I pass upstream. So this feels... fine?


__tests__/oauth-provider.test.ts
```diff
@@ -494,6 +494,30 @@ describe('OAuthProvider', () => {
...
+      // Create an authorization request with an invalid redirect URI
+      const invalidRedirectUri = 'https://attacker.example.com/callback';
+      const authRequest = createMockRequest(
+        `https://example.com/authorize?response_type=code&client_id=${clientId}` +
+          `&redirect_uri=${encodeURIComponent(invalidRedirectUri)}` +
+          `&scope=read%20write&state=xyz123`
+      );
+
+      // Mock parseAuthRequest to simulate our error handling
+      vi.spyOn(authRequest, 'formData').mockImplementation(() => {
```
kentonv, line 507: This spyOn looks wrong?

kentonv: Removed it, test still passes.

kentonv (approved):

PR #31, opened 2025-05-15
remove explicit imports from `@cloudflare/workers-types`

threepointone:
We expect people to be using cloudflare/workers-types anyway, and we want to match whatever version they're using (to avoid problems like https://github.com/cloudflare/ai/pull/161

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,10 +21,8 @@
     "prepublishOnly": "npm run build",
     "prettier": "prettier -w ."
   },
-  "dependencies": {
-    "@cloudflare/workers-types": "^4.20250311.0"
-  },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


kentonv (approved):
How exactly does typescript know to load the types though, if there's no import?

threepointone:
the root project loads up `@cloudflare/workers-types` like so https://github.com/cloudflare/workers-oauth-provider/blob/2c2c67c87ac64b7625c7725900083e9f00769948/tsconfig.json#L7 which makes them "globals". All Workers projects have this (or the newer autogenerated types via `wrangler types`)

kentonv:
Got it, thanks.

PR #40, opened 2025-06-04
Adding `customProps` to API

### LINKED ISSUE #36: FR: maintain ctx.props
I'm trying to access a custom header in order to conditionally load tools on an MCP server. 

I'm setting `ctx.props` in the `fetch` method, expecting to access it later in the `init` method.

However, it appears that the `ctx.props` set during `fetch` is being flushed or not persisted, as I'm unable to access the expected values in `init`.

__Steps to Reproduce:__

1. Set a custom header in the request.
2. In the fetch method, read this header and assign it to ctx.props.
3. Attempt to read ctx.props in init to control conditional tool logic.

__Expected Behavior:__

The value set in `ctx.props` during `fetch` should be accessible in `init`.

__Actual Behavior:__

`ctx.props` appears to be empty or reset by the time `init` is called post auth.

//cc @geelen (we have discussed about this)

geelen:
This lets a user add additional context per-request to their ApiHandler, by providing a `customProps` callback. The resulting object will be added to `ctx.props.customProps`:

```ts
new OAuthProvider({
  // ... other options ...
  customProps: async (request, env, ctx) => {
    // Will be added to ctx.props.customProps when invoking apiHandler
    return {
      customToken: request.headers.get('x-custom-token'),
      featureFlags: await env.FEATURES.parse(request.headers.get('x-feature-flags')),
    }
  }
})
```

This fixes #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,36 @@ new OAuthProvider({
 
 By default, the `onError` callback is set to ``({ status, code, description }) => console.warn(`OAuth error response: ${status} ${code} - ${description}`)``.
 
+## Passing custom or per-request props
+
+If your application needs to pass through context from each request to the ApiHandler, you can pass a `customProps` callback. This receives the same parameters (`request`, `env`, and `ctx`) as a standard Workers event handler, and lets you return arbitrary data that will be forwarded to your handler inside `ctx.props.customProps`.
+
+```ts
+new OAuthProvider({
+  // ... other options ...
+  customProps: async (request, env, ctx) => {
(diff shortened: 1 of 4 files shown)


geelen:
Ah this is probably another case of the Agents + Oauth combo hiding where the fix needs to happen. Agents does this: https://github.com/cloudflare/agents/blob/main/packages/agents/src/mcp/index.ts#L608

```js
await doStub._init(ctx.props);
```

So anything you add to `ctx.props` would have come right through. But yeah, potentially it's the McpAgent api that needs an API extension to allow converting headers to props instead, if it's already getting the raw request?

PR #52, opened 2025-07-24
Fix for the Missing Validation for ClientId

cnallam:
~https://github.com/cloudflare/workers-oauth-provider/security/advisories/GHSA-4pc9-x2fx-p7vj~

__(edited by @threepointone to strike out the advisory link since it's not relevant)__

**Original Request:** 
http://localhost:8787/authorize?response_type=code&client_id=k331WCMdRWmBxKxt&code_challenge=Jco6hZ5K7VFomiStpBRMo2VvUUNSBZQ9EnDZl0dBz3c&code_challenge_method=S256&redirect_uri=http%3A%2F%2Flocalhost%3A9882%2Foauth%2Fcallback&state=ce9ff539-17bf-40bf-900b-03c22960d47b

**Modified Request**:  Changed client_id and redirect_uri
http://localhost:8787/authorize?response_type=code&client_id=attackerId&code_challenge=Jco6hZ5K7VFomiStpBRMo2VvUUNSBZQ9EnDZl0dBz3c&code_challenge_method=S256&redirect_uri=http://www.google.com&state=ce9ff539-17bf-40bf-900b-03c22960d47b

After Successful authentication, token is redirected to www.google.com
image

**After Fix** : Catches the error: Error: Invalid client.

image

### DIFF
--- a/.changeset/sharp-apes-punch.md
+++ b/.changeset/sharp-apes-punch.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-oauth-provider": patch
+---
+
+Fix for the Missing Validation for ClientId
--- a/__tests__/oauth-provider.test.ts
+++ b/__tests__/oauth-provider.test.ts
@@ -511,6 +511,41 @@ describe('OAuthProvider', () => {
       expect(grants.keys.length).toBe(0);
     });
 
+    it('should reject authorization request with invalid client id', async () => {
+      // Create an authorization request with an invalid redirect URI
+      const invalidClientId = 'attackerClientId'
+      const authRequest = createMockRequest(
+        `https://example.com/authorize?response_type=code&client_id=${invalidClientId}` +
+          `&redirect_uri=${encodeURIComponent(redirectUri)}` +
+          `&scope=read%20write&state=xyz123`
+      );
+
+      // Expect the request to be rejected
+      await expect(oauthProvider.fetch(authRequest, mockEnv, mockCtx)).rejects.toThrow('Invalid client');
+
+      // Verify no grant was created
+      const grants = await mockEnv.OAUTH_KV.list({ prefix: 'grant:' });
+      expect(grants.keys.length).toBe(0);
+    });
+
+    it('should reject authorization request with invalid client id and redirect uri', async () => {
+      // Create an authorization request with an invalid redirect URI
+      const invalidRedirectUri = 'https://attacker.example.com/callback';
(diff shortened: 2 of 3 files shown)


kentonv:
While I agree that this is a bug, it is not a security vulnerability.

The vulnerability you linked was fixed in v0.0.5 (before the report was published). The bug addressed in this PR is _not_ a repeat of the vulnerability. The vulnerability was specifically that it was possible for an attacker to leverage authorization that had already been granted to the given client ID, by spoofing a request for that client ID, using a different redirect_uri. In this bug, you are specifying a client ID that doesn't exist, so obviously couldn't have any authorization granted to it already. Moreover, the auth code you create with this bug cannot be exchanged for a working token, as redeeming the auth code for a token requires authenticating using the client secret, but if there's no client registration at all, then there is obviously no client secret.

With all that said, I do agree that we shouldn't get to the point of issuing an auth code at all for an invalid client ID. But I don't think there's a security issue here.

Note that if there were a security issue -- or if there is a security issue here that I am not seeing -- then you should disclose it to us privately through HackerOne -- see [SECURITY.md](https://github.com/cloudflare/workers-oauth-provider/blob/main/SECURITY.md). Please do not report security issues using public bug reports or pull requests.

cnallam:
> While I agree that this is a bug, it is not a security vulnerability.
> 
> The vulnerability you linked was fixed in v0.0.5 (before the report was published). The bug addressed in this PR is _not_ a repeat of the vulnerability. The vulnerability was specifically that it was possible for an attacker to leverage authorization that had already been granted to the given client ID, by spoofing a request for that client ID, using a different redirect_uri. In this bug, you are specifying a client ID that doesn't exist, so obviously couldn't have any authorization granted to it already. Moreover, the auth code you create with this bug cannot be exchanged for a working token, as redeeming the auth code for a token requires authenticating using the client secret, but if there's no client registration at all, then there is obviously no client secret.
> 
> With all that said, I do agree that we shouldn't get to the point of issuing an auth code at all for an invalid client ID. But I don't think there's a security issue here.
> 
> Note that if there were a security issue -- or if there is a security issue here that I am not seeing -- then you should disclose it to us privately through HackerOne -- see [SECURITY.md](https://github.com/cloudflare/workers-oauth-provider/blob/main/SECURITY.md). Please do not report security issues using public bug reports or pull requests.

This is not a new vulnerability, there is another way to replicate the the fixed vulnerability "GHSA-4pc9-x2fx-p7vj". Just added the fix to check the clientId.

PR #62, opened 2025-08-07
token revocation endpoint support (fixes #49)

### LINKED ISSUE #49: revokeGrant not implemented in handleTokenRequest?
In the handleMetadataDiscovery the line 'revocation_endpoint: tokenEndpoint' tells all clients that they can send revocation requests to the same URL as the token endpoint.

The main router in the fetch method directs all requests for the tokenEndpoint to the handleTokenRequest function. This is the function that should contain the logic for both issuing tokens and revoking them. However, it completely lacks the revocation part.

It will return an unsupported_grant_type error.

However - I noticed we have a function called revokeGrant within OAuthHelpersImpl

This seems like it would work - however it is missing from handleTokenRequest

Is this an oversight?

whoiskatrin:
uses shared request parsing with parameter-based routing: `parseTokenEndpointRequest()` handles client authentication and body parsing once, then routes to either `handleRevocationRequest()` or `handleTokenRequest()` based on request parameters

claude code was using in this PR


src/oauth-provider.ts
```diff
@@ -1621,6 +1661,126 @@ class OAuthProviderImpl {
...
+          (await this.validateRefreshToken(token, userId, grantId, env));
+      }
+
+      // If we found a valid token, revoke the entire grant
+      if (isValidToken) {
+        await this.createOAuthHelpers(env).revokeGrant(grantId, userId);
+      }
+
+      return new Response('', { status: 200 });
+    } catch (error) {
```
kentonv, line 1727: I don't think we should catch exceptions here. An exception thrown in this code is almost certainly due to a bug in the code. We should let it propagate out so that it can b caught either by the top-level request handler of the application (which might e.g. report it to Sentry) or by tail workers.

I know that the RFC says that the server should return 200 even for invalid tokens, but errors caught here likely aren't due to invalid tokens, but rather bugs, and so returning a 500 would be appropriate.

src/oauth-provider.ts
```diff
@@ -1621,6 +1661,126 @@ class OAuthProviderImpl {
...
+      const tokenData = await env.OAUTH_KV.get(tokenKey, { type: 'json' });
+
+      if (!tokenData) {
+        return false;
+      }
+
+      // Check if token is expired
+      const now = Math.floor(Date.now() / 1000);
+      return tokenData.expiresAt >= now;
+    } catch {
```
kentonv, line 1754: I don't think we should catch exceptions and assume it means the token is invalid. An exception caught here almost certainly is due to some sort of bug in the code. Silently ignoring it and returning false means we won't ever find out about the bug -- we'll just have tokens that aren't correctly revoked.

kentonv (commented):
Was this written by hand or with AI? No problem using AI obviously, but if so the commits / PR description should say so, and ideally include the prompts.

whoiskatrin:
> Was this written by hand or with AI? No problem using AI obviously, but if so the commits / PR description should say so, and ideally include the prompts.

used detailed solution in #49 as the base for my implementation and fleshed out the full RFC 7009 requirements with Claude Code, wasn't really familiar with that

kentonv:
Please update the PR description to say that Claude Code was used.

(Meanwhile I just switched the repo settings to require squash+merge with PR title+description as the final commit title+description...)

kentonv (approved):

PR #67, opened 2025-08-08
test: granular CORS tests

ctjlewis:
Per our convo I added some more granular CORS tests but they are all passing. So, still debugging what is going on with my deployment in particular... But figured I'd submit these.


threepointone (approved):
Thank you!

PR #74, opened 2025-09-09
Add configurable refresh token TTL

### LINKED ISSUE #15: Consider options to make grants expire after some number of months, or prohibit refresh tokens etirely
I personally think that having grants expire is a bit user-hostile... like, suddenly my app breaks after a year, wtf?

But it really depends on the use case.

Also there should probably be an option to not even give out refresh tokens, for apps that only need access while the user is present, so can always request a new grant.

ghostwriternr:
## Summary
Adds configurable refresh token expiration through a new `refreshTokenTTL` option. Refresh tokens can now expire after a specified period, forcing users to reauthenticate.

## Features
- New `refreshTokenTTL` option in seconds (e.g., 3600 = 1 hour, 86400 = 1 day)
- Set to `0` to disable refresh tokens entirely (access token only mode)
- Omit or set to `undefined` for never-expiring refresh tokens (backward compatible)
- Per-grant TTL override via `tokenExchangeCallback` during authorization
- Automatic grant cleanup via KV expiration when refresh tokens expire

## Implementation
- Grants store `expiresAt` timestamp when TTL is configured
- KV storage uses absolute expiration timestamps for automatic cleanup
- Access tokens are clamped to not exceed refresh token's remaining lifetime
- Refresh token expiration is immutable after grant creation

## Usage
```typescript
// Global 30-day expiration for all refresh tokens
new OAuthProvider({
  refreshTokenTTL: 2592000 // 30 days
});

// Access tokens only, no refresh tokens
new OAuthProvider({
  refreshTokenTTL: 0
});

// Override TTL per grant during authorization
tokenExchangeCallback: async (options) => {
  if (options.grantType === 'authorization_code') {
    return { 
      refreshTokenTTL: 3600 // 1 hour for this specific grant
    };
  }
}
```

---
Claude Code was used in this PR for generating the initial version of the changes, as well as for writing all tests.
<details>
<summary>Initial prompt</summary>

> Customer _redacted_ wants users to reauthenticate after 1 hour. But we don't currently support a TTL for the refresh token, but we probably should.
>
> At present the refresh token doesn't expire. But you can call env.OAUTH_PROVIDER.revokeGrant() to revoke it. A possible work-around would be to put an expiration date in the props, and then check for expiration in each request. But adding support for expiring refresh tokens should be pretty easy. In fact we already have this issue for it, just never got around to it: https://github.com/cloudflare/workers-oauth-provider/issues/15 (That says "expire after some number of months" but there's no reason it couldn't also be "after an hour".)
>
> Use gh cli to view the issue, go through the existing codebase in detail, then come up with a plan to address this.
</details>

Fixes #15

### DIFF
--- a/src/oauth-provider.ts
+++ b/src/oauth-provider.ts
@@ -1705,8 +1705,6 @@ class OAuthProviderImpl {
     grantData.refreshTokenId = newRefreshTokenId;
     grantData.refreshTokenWrappedKey = newRefreshTokenWrappedKey;
 
-    // Note: The refresh token expiration (if any) remains unchanged during refresh
-
     // Save the updated grant with TTL if applicable
     await this.saveGrantWithTTL(env, grantKey, grantData, now);
 


README.md
```diff
@@ -261,12 +266,56 @@ The callback can:
 - Return only `accessTokenProps` to update just the current access token
 - Return only `newProps` to update both the grant and access token (the access token inherits these props)
 - Return `accessTokenTTL` to override the default TTL for this specific access token
+- Return `refreshTokenTTL` to override the default TTL for this specific refresh token
 - Return nothing to keep the original props unchanged
 
 The `accessTokenTTL` override is particularly useful when the application is also an OAuth client to another service and wants to match its access token TTL to the upstream access token TTL. This helps prevent situations where the downstream token is still valid but the upstream token has expired.
 
+The `refreshTokenTTL` override allows you to set different expiration times for refresh tokens based on user type, client application, or other factors. For example, you might want shorter-lived refresh tokens for sensitive operations or longer-lived ones for trusted applications.
```
kentonv, line 274: I think Claude is being too verbose here. I'd delete this line and all the changes below this point. The meaning of this setting is obvious enough and doesn't need all this exposition.

README.md
```diff
@@ -87,7 +87,12 @@ export default new OAuthProvider({
   // Note: Creating public clients via the OAuthHelpers.createClient() method
   // is always allowed regardless of this setting.
   // Defaults to false.
-  disallowPublicClientRegistration: false
+  disallowPublicClientRegistration: false,
+
+  // Optional: Time-to-live for refresh tokens in seconds.
```
kentonv, line 92: We should also have an option for "no refresh token" (only access token). Maybe set this to zero?

src/oauth-provider.ts
```diff
@@ -475,6 +488,11 @@ export interface Grant {
    */
   createdAt: number;
 
+  /**
+   * Unix timestamp when the refresh token expires (if TTL is configured)
+   */
+  refreshTokenExpiresAt?: number;
```
kentonv, line 494: When the refresh token expires, the grant itself effectively expires, so I think this can just be called `expiresAt`.

src/oauth-provider.ts
```diff
@@ -643,6 +661,11 @@ export interface GrantSummary {
    * Unix timestamp when the grant was created
    */
   createdAt: number;
+
+  /**
+   * Unix timestamp when the refresh token expires (if TTL is configured)
+   */
+  refreshTokenExpiresAt?: number;
```
kentonv, line 668: `expiresAt`

src/oauth-provider.ts
```diff
@@ -1391,6 +1421,9 @@ class OAuthProviderImpl {
     const now = Math.floor(Date.now() / 1000);
     const accessTokenExpiresAt = now + accessTokenTTL;
 
+    // Calculate refresh token expiration time if TTL is configured
+    const refreshTokenExpiresAt = refreshTokenTTL ? now + refreshTokenTTL : undefined;
```
kentonv, line 1425: Technically this mishandles `refreshTokenTTL = 0`. Needs to explicitly check `refreshTokenTTL === undefined`.

src/oauth-provider.ts
```diff
@@ -1622,6 +1678,13 @@ class OAuthProviderImpl {
     grantData.refreshTokenId = newRefreshTokenId;
     grantData.refreshTokenWrappedKey = newRefreshTokenWrappedKey;
 
+    // Update refresh token expiration if a new TTL was provided by callback, otherwise keep existing expiration
+    if (refreshTokenTTLUpdated) {
+      // If TTL is 0 or null, remove expiration; otherwise set new expiration
```
kentonv, line 1683: Treating `null` as "no expiration" is OK, but treating 0 as "no expiration" seems dangerous. If anything it should be treated as "expires immediately" (though I'm also OK with it meaning "don't issue a refresh token").

kentonv (commented):
* When an access token is created from a refresh token, if that refresh token is imminently expiring, we should clamp the access token's expiration to that of the refresh token.
* I think we should support setting refreshTokenTTL to zero to mean "don't issue a refresh token". In this case you only get an access token (with the TTL as configured for access tokens).

kentonv:
Also: Please make sure the commit description mentions that you used AI and ideally include the prompts.

src/oauth-provider.ts
ghostwriternr: Agreed, that is much cleaner

ghostwriternr:
Thanks for the comments, @kentonv! I've resolved everything, especially changing the interpretation of `refreshTokenTTL: 0` to consistently be "don't issue refresh tokens". The part that confused me the most, tbh, is ensuring the guarantees of having a grant retain 2 valid refresh tokens and being resilient to network failures is retained. Sufficient tests have been added (this, thanks to claude code again) to give me confidence the solution works, but I'd like your opinion on it because it doesn't _feel clean_ even if it is correct (if at all it is).

src/oauth-provider.ts
```diff
@@ -1650,43 +1679,82 @@ class OAuthProviderImpl {
...
+      }
+    }
 
     // Update refresh token expiration if a new TTL was provided by callback, otherwise keep existing expiration
     if (refreshTokenTTLUpdated) {
-      // If TTL is 0 or null, remove expiration; otherwise set new expiration
-      grantData.refreshTokenExpiresAt = refreshTokenTTL ? now + refreshTokenTTL : undefined;
+      if (isFinalRefresh) {
+        // For final refresh, we don't update expiration - the current token keeps its existing expiration
+        // This allows the client to retry this final refresh if needed
```
ghostwriternr, line 1746: Is this behaviour a fair expectation? From user pov

src/oauth-provider.ts
```diff
@@ -1650,43 +1679,82 @@ class OAuthProviderImpl {
...
+      //
+      // We update the wrapped key for whichever token was used (current or previous) with the
+      // key that successfully authenticated this request. This ensures retries work correctly,
+      // especially when props change (which changes the grant's encryption key).
+      if (isCurrentToken) {
+        grantData.refreshTokenWrappedKey = wrappedKeyToUse;
+      } else {
+        grantData.previousRefreshTokenWrappedKey = wrappedKeyToUse;
+      }
+    }
```
ghostwriternr, line 1740: This is the part I'm worried about if I'm honest (but technically works)

src/oauth-provider.ts
kentonv: No, I don't think it is.

`refreshTokenTTL = 0` is supposed to mean "don't issue a refresh token". But here we _already_ issued one, but on a refresh operation, the callback returned this. I guess the most sensible thing to do would be to delete the existing refresh token.

The grant record needs to continue existing, though, until all access tokens are gone. So I guess you'd want to set its TTL in KV to match the access token.

This all seems like a lot of work to support something nobody asked for, though. Maybe we just shouldn't support changing the TTL after the grant is created?

src/oauth-provider.ts
```diff
@@ -2068,6 +2140,21 @@ class OAuthProviderImpl {
...
+  /**
+   * Saves a grant to KV with appropriate TTL based on expiration
+   * @param env - The environment bindings
+   * @param grantKey - The KV key for the grant
+   * @param grantData - The grant data to save
+   * @param now - Current timestamp in seconds
+   */
+  private async saveGrantWithTTL(env: any, grantKey: string, grantData: Grant, now: number): Promise<void> {
+    // Calculate TTL: undefined if no expiration or already expired (shouldn't happen but handled gracefully)
+    const grantTTL = grantData.expiresAt && grantData.expiresAt > now ? grantData.expiresAt - now : undefined;
```
kentonv, line 2152: I think if `grantData.expiresAt` is before `now` -- i.e. the grant is already expired -- we probably want to delete the record, rather than save it with no TTL?

src/oauth-provider.ts
```diff
@@ -2068,6 +2140,21 @@ class OAuthProviderImpl {
...
+    // Add TTL to KV options if present, otherwise grant persists indefinitely
+    const kvOptions = grantTTL ? { expirationTtl: grantTTL } : {};
```
kentonv, line 2154: This should probably use the `expiration` option, which takes an absolute date, rather than `expirationTtl`, which is relative. Then the computation of `grantTTL` above isn't needed. This probably also takes care of the edge case where the grant is already expired.

PR #78, opened 2025-09-11
 Use rejection sampling to avoid bias in `generateRandomString()`

### LINKED ISSUE #42: Biased output from getRandomString()
The `getRandomString()` implementation produces biased output. The characters A-H are more likely to be produced than any others (about 1.5% more likely). This is due to the use of `% characters.length` to reduce each byte to a value in the alphabet, which introduces bias. You should either use rejection sampling or extra random bits (at least 64 extra random bits) to reduce the bias.

This bug reduces the effective entropy of all of the tokens generated by this library, although unlikely to a level that poses a real security threat (hence why disclosing here and not via the bug bounty).

rc4:
Fix #42 by rejecting outputs >= `characters.length * 4` (248). 

We initially generate 10% more random bytes than requested. If we expect to reject 3.1% of them; only one call to `crypto.getRandomValues()` should be required in >99% of cases. 

If not, we get fresh values and go again until we've satisfied the requested length.

Let me know if there are any questions/concerns/anything I need to do on my end - I dug through the docs and it seemed pretty straightforward to contribute. :smile:

### DIFF
--- a/.changeset/tasty-baboons-ring.md
+++ b/.changeset/tasty-baboons-ring.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-oauth-provider": patch
+---
+
+ Use rejection sampling to avoid bias in `generateRandomString()`


kentonv:
Hmm, thanks for taking a pass at it, but I was sort of thinking we should just add two characters to the alphabet instead. If we add `_` and `-` then we have an even power-of-two alphabet which eliminates the bias.

rc4:
Even easier! I am perfectly happy to do that instead, but it seemed that you might have had a few reservations about taking that approach so I figured this would be a fun little exercise to try. If not that simplifies the implementation a good bit :slightly_smiling_face:

PR #81, opened 2025-09-11
Add resolveExternalToken to support external token auth flows

deathbyknowledge:
Adds `resolveExternalToken` to support auth for external tokens. The callback only runs IF internal auth check fails.

E.g.  a canonical OAuth server is used by multiple services, allowing server-server communication with the same token.

Example usage:
```typescript
const provider = new OAuthProvider({
  apiRoute: '/api',
  apiHandler,
  defaultHandler,
  authorizeEndpoint: '/authorize', // Still used for internal auth flow
  tokenEndpoint: '/oauth/token',
  resolveExternalToken: async ({ token, request, env }) => {
    // Say we primarily use JWKS for auth
    try {
      const JWKS = jose.createRemoteJWKSet(new URL(env.IDP_JWKS_URI));
      const { payload } = await jose.jwtVerify(token, JWKS, {
        issuer: env.IDP_ISSUER,  // You should set this in your ENV
        audience: env.API_AUDIENCE, // This too
        algorithms: ['RS256','RS384','RS512','ES256','ES384','ES512'],
      });
  
      // Anything you might want to add to your props
      const props = {
        email: (payload as any).email ?? null,
        name: (payload as any).name ?? null,
      };
  
      return { props };
    } catch { } 

    // Try RFC 7662 introspection for opaque tokens as a fallback
    // (just an example of how you could implement this)
    
    // These must also be set in your env as vars/secrets
    if (env.INTROSPECTION_URL && env.INTROSPECTION_ID && env.INTROSPECTION_SECRET) {
      const body = new URLSearchParams({ token, token_type_hint: 'access_token' });
      const resp = await fetch(env.INTROSPECTION_URL, {
        method: 'POST',
        headers: {
          'content-type': 'application/x-www-form-urlencoded',
          Authorization:
            'Basic ' +
            btoa(
              `${encodeURIComponent(env.INTROSPECTION_ID)}:${encodeURIComponent(env.INTROSPECTION_SECRET)}`
            ),
        },
        body,
      });
      if (!resp.ok) return null;
  
      const data = await resp.json();
      if (!data?.active) return null;
  
      // Basic issuer/audience/time checks
      if (data.iss && data.iss !== env.IDP_ISSUER) return null;
      const audOk = Array.isArray(data.aud)
        ? data.aud.includes(env.API_AUDIENCE)
        : !data.aud || data.aud === env.API_AUDIENCE;
      if (!audOk) return null;
      const now = Math.floor(Date.now() / 1000);
      if ((data.nbf && now < data.nbf - 5) || (data.exp && now > data.exp + 5)) return null;
  
      const props = {
        email: data.email ?? null,
        name: data.name ?? null,
        // again, we can set anything here
      };
  
      return { props };
    }
  
    // Neither JWT nor introspection validated the token
    return null;
  }
 });
```

### DIFF
--- a/src/oauth-provider.ts
+++ b/src/oauth-provider.ts
@@ -230,7 +230,7 @@ export interface OAuthProviderOptions {
    * For example, if a request includes an authenticated token from a different OAuth authentication server,
    * the callback can be used to authenticate it and set the context props through it.
    *
-   * The callback can optionally return props values that will passed-through to the apiHandlers.  
+   * The callback can optionally return props values that will passed-through to the apiHandlers.
    * The callback can return `null` to signal resolution failure.
    */
   resolveExternalToken?: (input: ResolveExternalTokenInput) => Promise<ResolveExternalTokenResult | null>;


kentonv:
Hmm, what's the flow for that exactly? How does someone end up with an API key for your normal API, but want to plug it into MCP? Wouldn't an MCP client always initiate a new OAuth request to the MCP endpoint itself?

Not saying it's wrong, I'm just trying to understand what I'm looking at.

kentonv (approved):

PR #92, opened 2025-10-16
fix: open redirect vulnerability in completeAuthorization

roerohan:
## Summary of Fix
This change addresses an open redirect vulnerability in the `completeAuthorization` method.

## The Vulnerability
The `completeAuthorization` method previously trusted the `redirect_uri` provided within the `AuthRequest` object without re-validating it. An application using this library could inadvertently pass an `AuthRequest` object constructed from untrusted data (e.g., a state parameter from the URL). We were trusting the implementor of the library to ensure that a validated `redirect_uri` is passed in `completeAuthorization`.

For example, if the implementor stores the `redirect_uri` in the "?code=...&state=..." URL string, it gives an attacker a vector to modify the `redirect_uri`. An attacker can craft a malicious authorization link with a modified state parameter containing an attacker-controlled `redirect_uri`. If a victim authenticated through this link, the authorization code would be sent to the attacker's server, leading to a potential account takeover.

## The Solution
This fix adds a validation block at the beginning of the `completeAuthorization` method. Before any grant is created, the method now:
1. Looks up the client using the `clientId` from the request.
2. Verifies that the `redirectUri` from the request is present in the client's list of registered `redirectUris`.

This change ensures that the `redirect_uri` is always validated before use, making the library secure by default and protecting all downstream applications from this vulnerability.

### DIFF
--- a/.changeset/curly-pumpkins-smile.md
+++ b/.changeset/curly-pumpkins-smile.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-oauth-provider": patch
+---
+
+fix: open redirect vulnerability in completeAuthorization


kentonv:
This validation is reasonable to do as a defensive measure, but note that this doesn't really solve any "open redirect" issue because an attacker can just use dynamic client registration to register a client with whatever redirect_uri they want.

Moreover, it should not be possible for an attacker to inject a bogus request_uri between `parseAuthRequest()` and `completeAuthorization()`. It is the responsibility of the app not to pass the parsed request out of its custody such that it could be modified or injected by an attacker between these two steps.

mschwarzl:
Hi, as you both correctly discussed as long as you can arbitrarily register a client, the mitigation won't help against the phishing issue. That's why I was also asking to do csrf checking on the application/client side as well.

Preventing these kind of phishing attacks where a victim just copy/pastes an URL and it directly causes an action in the logged in env. I agree that the suggested fix would still improve the current situation. 

Moreover, I'd like to see some recommendations / sample client usage that users/customers could refer to.

PR #96, opened 2025-10-21
feat: validate dynamic client registration

roerohan:
This change introduces a new `validateClientRegistration` callback to the `OAuthProvider` options, enhancing the security of the dynamic client registration feature.

## The Problem

Previously, when the `clientRegistrationEndpoint` was enabled, the library allowed any client to register without restriction. This posed a security risk, as it could allow malicious or unauthorized clients to be registered with the OAuth provider.

## The Solution
To address this, I have added a new optional callback function, `validateClientRegistration`, to the `OAuthProviderOptions`. This function is executed before a new client is registered and allows developers to implement custom validation logic.

The callback receives the `clientMetadata` and the original request as arguments and should return a `boolean` value indicating whether the registration is allowed. If the function returns `false`, the registration is denied with a 403 Forbidden error.

I have also updated the `README.md` to:
- Add a new "Security" section.
- Document the importance of CSRF prevention using the state parameter.
- Provide a detailed explanation and a code example for using the new `validateClientRegistration` callback to secure the dynamic client registration endpoint.

This change empowers developers to enforce their own policies on client registration, such as allowlisting specific domains for redirect URIs or requiring an authentication token for the registration request.

### DIFF
--- a/.changeset/itchy-carrots-dig.md
+++ b/.changeset/itchy-carrots-dig.md
@@ -3,3 +3,4 @@
 ---
 
 feat: add a function to validate client registration
+fix: make request the first parameter


roerohan:
> The whole point of DCR is to let arbitrary clients register with the service. An MCP server, for example, wants to allow _any_ agent to use it, including agents the service owner hasn't heard of.
> 
> So what validation can they possibly do?
> 
> If you have a server that _doesn't_ want to allow everyone to register, then you should turn of DCR, and instead issue client credentials to the specific clients that you want to allow, or implement an alternative registration mechanism that matches your business needs.

I see your point - I was under the assumption that DCR could still be useful for organizations that want to enable dynamic registration for clients within their own domain. For example, an internal MCP server might allow DCR for all clients hosted on the same domain, or maybe just for trusted internal tools and editors. In that kind of controlled environment, DCR could still make sense without opening it up to arbitrary external clients.

PR #112, opened 2025-12-03
feat: add Client ID Metadata Document (CIMD) support

mattzcarey:
CIMD support allows clients to use HTTPS URLs as client_id values that point to metadata documents. 

When a client_id is an HTTPS URL with a non-root path, the provider fetches and validates the metadata document instead of looking up in KV storage. Added validation to ensure client_id in the document matches the URL and redirect_uris are present.

matches the new authorization spec for MCP

https://modelcontextprotocol.io/specification/draft/basic/authorization

### DIFF
--- a/src/oauth-provider.ts
+++ b/src/oauth-provider.ts
@@ -2029,7 +2029,8 @@ class OAuthProviderImpl {
     }
 
     // Get token endpoint auth method, default to client_secret_basic
-    const authMethod = OAuthProviderImpl.validateStringField(clientMetadata.token_endpoint_auth_method) || 'client_secret_basic';
+    const authMethod =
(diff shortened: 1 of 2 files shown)


src/oauth-provider.ts
```diff
@@ -2290,19 +2293,284 @@ class OAuthProviderImpl {
...
   }
 
+  /**
+   * Checks if a client_id is a valid CIMD URL
+   * Must be HTTPS with a non-root pathname
+   */
+  private isClientMetadataUrl(clientId: string): boolean {
+    try {
+      const url = new URL(clientId);
+      return url.protocol === 'https:' && url.pathname !== '/';
```
pcarleton: this is correct to reject here, but you may want a louder failure if someone's passed this situation, since i think with this it will fail kind of oddly (like will it fall back to DCR?).  i imagine it'll be rare, so fine to cross that bridge if it comes up.

PR #114, opened 2025-12-05
Token introspection

### LINKED ISSUE #72: Create an introspection endpoint
Here's what I'm doing right now (I've got React Router set up in my worker alongside `@cloudflare/workers-oauth-provider`):

```ts
import { invariantResponse } from '@epic-web/invariant'
import { type Token } from '#types/helpers'
import { type Route } from './+types/introspection.ts'

export async function action({ request, context }: Route.LoaderArgs) {
	const token = (await request.formData()).get('token')?.toString()
	invariantResponse(token, 'invalid_request')

	const parts = token.split(':')
	if (parts.length !== 3) return { active: false }

	const [userId, grantId] = parts
	const tokenId = await generateTokenId(token)
	const tokenKey = `token:${userId}:${grantId}:${tokenId}`

	const tokenData = await context.cloudflare.env.OAUTH_KV.get(tokenKey, {
		type: 'json',
	})

	if (!tokenData) return { active: false }

	const info = tokenData as Token

	if (info.expiresAt * 1000 < Date.now()) {
		return { active: false }
	}

	return {
		active: true,
		client_id: info.grant.clientId,
		scope: info.grant.scope.join(' '),
		sub: info.userId,
		exp: info.expiresAt,
		iat: info.createdAt,
	}
}

// copied from @cloudflare/workers-oauth-provider
async function generateTokenId(token: string) {
	const encoder = new TextEncoder()
	const data = encoder.encode(token)
	const hashBuffer = await crypto.subtle.digest('SHA-256', data)
	const hashArray = Array.from(new Uint8Array(hashBuffer))
	const hashHex = hashArray.map((b) => b.toString(16).padStart(2, '0')).join('')
	return hashHex
}
```

It technically works, but I would rather not have to rely on undocumented `tokenKey` formats and copy/pasted code 😅

Even better if it could work like other auth providers so I could easily swap out this for [WorkOS](https://workos.com/docs/authkit/mcp/integrating/token-verification)/etc if the code in my resource server were set up like this:

```ts
import { jwtVerify, createRemoteJWKSet } from 'jose';

const JWKS = createRemoteJWKSet(new URL('https://whatever.authkit.app/oauth2/jwks'));

// ...

async function verifyToken(req: Request) {
	const token = req.headers.get('authorization')?.match(/^Bearer (.+)$/)?.[1];
	// ...

	const { payload } = await jwtVerify(token, JWKS, {
		issuer: 'https://whatever.authkit.app',
	});

	// Use access token claims to populate request context.
	// i.e. `const userId = payload.sub;`

	// ...
}
```

DeanMauro:
This PR adds a method `decodeToken` that retrieves a granted access token from the KV and returns the user-defined props attached to it.

This permits token decoding outside of a fetch call, e.g. an RPC call from another worker.

## Example
_Consider a standalone auth server shared by a number of micro-services_

**Worker A (Auth Server)**
```
const oauthProvider = new OAuthProvider({...});

export default class extends WorkerEntrypoint {
	async fetch(request: Request) {
		return oauthProvider.fetch(request, this.env, this.ctx);
	}

	public async readToken(token: string): Promise<string> {
		const tokenSummary = await env.OAUTH_PROVIDER.decodeToken(token, this.env);
		if (!tokenSummary) throw new Error("Invalid or expired token");

		return tokenSummary;
	}
}
```

**Worker B (API Server)**
```
const token = <caller provided token>;
const data = await this.env.WORKER_A.readToken(token);

// Bound worker now has caller's identity and can check scopes, metadata, etc.
```

Resolves #72

### DIFF
--- a/.changeset/late-points-jog.md
+++ b/.changeset/late-points-jog.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+adds a method `decodeToken` that retrieves a granted access token from the KV and returns the user-defined props attached to it. This permits token decoding outside of a fetch call, e.g. an RPC call from another worker.


DeanMauro:
I broadened the method, returning all token data rather than just the decrypted props, so it can resolve #72.

src/oauth-provider.ts
```diff
@@ -682,6 +689,19 @@ export interface Token {
   };
 }
 
+/**
+ * Token record with decrypted properties
+ * Derived from Token but with wrappedEncryptionKey removed and encryptedProps replaced with metadata
+ */
+export type TokenSummary<T = any> = Omit<Token, 'wrappedEncryptionKey' | 'grant'> & {
```
kentonv: Let's actually write out this type rather than declare it as `Token` with fields omitted... otherwise it's too easy to add something to `Token` that isn't appropriate for `TokenSummary` and not realize it will be exposed.

src/oauth-provider.ts
```diff
@@ -345,6 +345,13 @@ export interface OAuthHelpers {
    * @returns A Promise resolving when the revocation is confirmed.
    */
   revokeGrant(grantId: string, userId: string): Promise<void>;
+
+  /**
+   * Decodes a token and returns token data with decrypted props
+   * @param token - The token
+   * @returns Promise resolving to token data with decrypted props, or null if token is invalid
+   */
+  decodeToken<T = any>(token: string): Promise<TokenSummary<T> | null>;
```
kentonv: Bikeshedding: Let's call it `unwrapToken`? It's not really "decoding" anything.

src/oauth-provider.ts
```diff
@@ -1010,6 +1048,55 @@ class OAuthProviderImpl {
...
+      userId: tokenData.userId,
+      createdAt: tokenData.createdAt,
+      expiresAt: tokenData.expiresAt,
+      audience: tokenData.audience,
+      grant: {
+        clientId: grant.clientId,
+        scope: grant.scope,
+        props: decryptedProps as T,
+      },
+    };
```
kentcdodds: Should this not include an `active` property as required by the OAuth 2.0 Token Introspection spec?

https://datatracker.ietf.org/doc/html/rfc7662#section-2.2

More generally, this response doesn't appear to be spec-compliant. Several required or recommended fields are either missing (active) or represented differently (exp, iat, sub, etc.), and some fields here don’t map cleanly to the introspection response at all.

Is this endpoint intentionally not an RFC 7662 introspection response (i.e. an internal representation), or is partial compliance acceptable here? If it’s intended to be an introspection endpoint, I think we should either align more closely with the spec or document the deviations explicitly.

DeanMauro: Yeah I didn't add the endpoint, but the `unwrapToken` method can sit behind a custom route and do the same. Main goal was to surface a method that could read tokens without having to look through the code and copy the "generateTokenId" machinery.
Can implement a proper RFC 7662 endpoint in another PR too.

PR #120, opened 2025-12-15
Token Exchange (RFC 8693)

### LINKED ISSUE #116: Support for OAuth 2.0 Token Exchange (RFC 8693)
I'd like to add support for the [OAuth 2.0 Token Exchange specification (RFC 8693)](https://datatracker.ietf.org/doc/html/rfc8693).
Token Exchange allows clients or backend services to exchange an existing access token for a new one with modified characteristics like:

- A narrowed set of scopes (downscoping)
- A different target audience (resource server)
- A shorter TTL
- (Optionally) an “actor token” representing service-to-service delegation

### Use Cases
- Public clients that want to obtain short-lived, least-privilege tokens for specific flows
- Backends minting temporary delegated tokens for downstream APIs on behalf of a user
- Federated environments that need audience-restricted tokens between multiple microservices

### Why?
Right now there’s no clean way to mint derived tokens. A client would need to go through the authorization flow a second time for a new token with narrower scopes.

#### 1. Expose `/token/exchange` as a standard endpoint (RFC 8693)
The endpoint takes in a subject_token (the user's current issued token) and optionally new scope or audience.

Example Request:
```
POST /token/exchange
Content-Type: application/x-www-form-urlencoded
Authorization: Basic <client creds>

grant_type=urn:ietf:params:oauth:grant-type:token-exchange&
subject_token=<access_token>&
subject_token_type=urn:ietf:params:oauth:token-type:access_token&
scope=user:read
```

#### 2. Add an new `exchangeToken` method to `OAuthHelpers`

```
const shortToken = await exchangeToken({
  subjectToken: accessToken,
  scope: ["user:read"],            // Optional
  aud: "serviceB"                     // Optional
  ttl: 60                                     // Optional (have a default built in)
});
```

#### 3. Verify
`exchangeToken` first uses `unwrapToken` from #114 to validate the subject_token that was passed in and pull its grant info.
It then validates that the requested changes are allowed by the original grant:
```
if (!requestedScopes.every(scope => tokenSummary.grant.scope.includes(scope))) {
  return createErrorResponse("invalid_scope", "Requested scope exceeds parent grant");
}
```

#### 4. Mint new token
If the request is kosher, mint a new token in the same way `handleAuthorizationCodeGrant` does.

### For Maintainers
@kentonv @mattzcarey @threepointone 

I'm happy to implement. Just want to get alignment before opening a PR.

DeanMauro:
Resolves #116 (see issue description)

### DIFF
--- a/.changeset/fuzzy-clouds-attend.md
+++ b/.changeset/fuzzy-clouds-attend.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Add OAuth 2.0 Token Exchange (RFC 8693) support. Clients can exchange an existing access token for a new one with narrowed scopes, a different audience, or a shorter TTL — without requiring the user to re-authorize. Gated behind the `allowTokenExchangeGrant` option (default `false`). Also adds scope downscoping (RFC 6749 Section 3.3) to authorization code and refresh token flows.
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "typecheck": "tsc",
     "test": "vitest run",
     "test:watch": "vitest",
-    "prepare": "npm run build",
     "prepublishOnly": "npm run build",
     "prettier": "prettier -w ."
   },


mattzcarey:
prelim review generated with me and Claude. I won't get to landing this during this week but please have a quick look at the below :) thanks.  

### Code review

Found 2 issues:

1. **Resource privilege escalation in token exchange** - When `requestedResource` is provided but `grantData.resource` is undefined, the validation at line 2158 is skipped, but `newAudience` is still set to the requested resource at line 2175. This allows a client to add audience/resource claims that were never authorized in the original grant. The authorization code and refresh token flows have similar structure but this is particularly concerning for token exchange which is specifically designed for downscoping.

https://github.com/cloudflare/workers-oauth-provider/blob/f4215e8cfcbc06a53b148c23e9e5adeec54e4017/src/oauth-provider.ts#L2155-L2176

2. **Token exchange callback receives token's scopes instead of grant's scopes** - The authorization code flow (line 1674) and refresh token flow pass `grantData.scope` to the callback, but the token exchange flow (line 2216) passes `tokenSummary.grant.scope`. If the subject token was created with downscoped permissions, the callback will receive those downscoped scopes rather than the original grant scopes, which is inconsistent with other flows and could affect authorization policy decisions.

https://github.com/cloudflare/workers-oauth-provider/blob/f4215e8cfcbc06a53b148c23e9e5adeec54e4017/src/oauth-provider.ts#L2212-L2219

DeanMauro:
@mattzcarey resurfacing this one in case you have time to test it out

mattzcarey (approved):

PR #144, opened 2026-02-23
feat: revoke existing grants on re-authorization by default

### LINKED ISSUE #34: No provision to restart the authorization flow if the oauth server to the worker oauth client does not provide refresh tokens
I am using this for tokenExchangeCallback. The problem I am facing is after I throw an error with invalid_grant to signify that the grant has expired in a refresh_token grantType call, it does call the /authorize endpoint and the token is fetched again from the oauth server using the original flow. But the props are not updated upon calling the completeAuthorization() function from OAUTH_PROVIDER, it still uses the old props. Any help here would be appreciated.
` tokenExchangeCallback: async (options) => {
        console.log("tokenExchangeCallback", options);
        if (options.grantType === "refresh_token") {
                if (options.grantType === "refresh_token") {
                  const { accessToken, tokenExpiryInSec, instanceUrl } = options.props;
                  console.log("options", options);
                  const now = Date.now();
              
                  // access token is expired — fetch a new one
              
                  const url = `${instanceUrl}/access/token/fetch?validity_time_in_sec={timeInSec}`;
              
                  const response = await fetch(url, {
                    method: "GET",
                    headers: {
                      Authorization: `Bearer ${accessToken}`, // old token (may still be valid)
                      Accept: "application/json",
                    },
                  });
              
                  if (!response.ok) {
                    
                    console.error("Failed to fetch new access token:", await response.text());
              
                    // Don't issue new Cloudflare token — force user to reauth
                    throw new Error(JSON.stringify({
                      error: "invalid_grant",
                      error_description: "access token expired. Please reauthenticate."
                    }));
                  }
              
                  const data = await response.json();
                  const newToken = data.data?.token;
                  const newExpiry = data.data?.expiration_time_in_millis;
              
              
                  return {
                    accessTokenProps: {
                      ...options.props,
                      accessToken: newToken,
                      tokenExpiryInSec: newExpiry
                    },
                    newProps: {
                      ...options.props,
                      accessToken: newToken,
                      tokenExpiryInSec: newExpiry
                    },
                    accessTokenTTL:120
                  };
              };
        }
        // fallback to default behavior for other grant types
        return;
    }, `

mattzcarey:
## Summary

Two related grant-security improvements, shipped as one body of work:

1. **Revoke existing grants on re-authorization** — When a user goes through the OAuth flow again for the same client, all previous grants (and their tokens) for that user+client are now revoked after the new grant is stored. This fixes infinite re-auth loops caused by stale props in old grants (issue #34).

2. **Revoke tokens on authorization code reuse (RFC 6749 §10.5)** — When an authorization code that has already been exchanged is presented again, the provider now revokes the grant and all tokens from the original exchange. This prevents authorization code replay attacks. (Originally PR #157 by @rc4, squashed into this PR.)

## Breaking behavior change

`completeAuthorization()` now defaults `revokeExistingGrants` to `true`. Previously, re-authorizing the same user+client created an additional grant, leaving old tokens valid. Now, old grants are revoked by default. This is an edge case but it could be hit by some MCP clients.

**If your application relies on multiple concurrent grants per user+client**, set `revokeExistingGrants: false` to preserve the old behavior:

```typescript
await oauthProvider.completeAuthorization({
  request: authRequest,
  userId: user.id,
  metadata: { ... },
  scope: ['read', 'write'],
  props: { ... },
  revokeExistingGrants: false,
});
```

MCP servers should not need to preserve this so setting the default to true here but allowing opt out for developers who need it. 

### `src/oauth-provider.ts`

- Added `revokeExistingGrants?: boolean` option to `CompleteAuthorizationOptions` (defaults to `true`)
- Before creating the new grant, collects all existing grant IDs for the same `userId + clientId` (with cursor-based pagination for >1000 grants)
- After the new grant is stored, revokes old grants in parallel via `Promise.all`
- Revocation is **best-effort** (wrapped in try/catch) — if it fails, the new grant is already stored and the authorization response still succeeds
- Both implicit flow and authorization code flow paths are covered
- Added auth code reuse detection: when `authCodeId` is absent on a grant, calls `revokeGrant()` (best-effort) and returns `invalid_grant`

### `__tests__/reauth-grant-revocation.test.ts` (new, 5 tests)

- Demonstrates stale props persist when `revokeExistingGrants: false` (the bug)
- Demonstrates the infinite re-auth loop scenario with `tokenExchangeCallback` throwing `invalid_grant`
- Old access tokens return 401 after re-authorization (fix works)
- Old refresh tokens return 400 after re-authorization (fix works)
- Other clients' grants are unaffected

### `__tests__/oauth-provider.test.ts` (1 new test)

- Auth code reuse: second exchange returns `invalid_grant`, all tokens and the grant are revoked

### `.changeset/fix-revoke-existing-grants.md`

- Minor version bump (default behavior change)

## RFC / spec alignment

- **RFC 6749 §10.5**: Authorization code reuse detection and token revocation — now implemented
- **[Grant Management for OAuth 2.0](https://openid.net/specs/fapi-grant-management-01.html)** (OpenID FAPI draft): Defines a `replace` action that does exactly what `revokeExistingGrants` does
- Most production OAuth servers (Auth0, Okta) revoke old refresh tokens on re-authorization as standard practice
- Neither OAuth 2.0 (RFC 6749), OAuth 2.1, nor the MCP spec mandate or prohibit revoking old grants on re-authorization — this is a defensible server-side policy choice

## Test plan

- [x] 222 tests pass (216 existing + 5 re-auth + 1 auth code reuse, no regressions)
- [x] Formatting passes (`prettier --check`)
- [x] E2E validation with MCP Inspector + real `agents` SDK MCP server

Fixes #34 | Includes #157

### DIFF
--- a/__tests__/oauth-provider.test.ts
+++ b/__tests__/oauth-provider.test.ts
@@ -6951,7 +6951,7 @@ describe('OAuthProvider', () => {
     // --- Helpers for the OAuth dance ---
 
     async function registerClient(
-      provider: OAuthProvider<any>,
+      provider: OAuthProvider<TestEnv>,
       env: any,
       ctx: MockExecutionContext
     ): Promise<{ clientId: string; clientSecret: string }> {
@@ -6974,7 +6974,7 @@ describe('OAuthProvider', () => {
     }
 
     async function authorizeAndGetCode(
-      provider: OAuthProvider<any>,
+      provider: OAuthProvider<TestEnv>,
       env: any,
       ctx: MockExecutionContext,
       clientId: string
@@ -6990,7 +6990,7 @@ describe('OAuthProvider', () => {
     }
 
     async function exchangeCodeForTokens(
-      provider: OAuthProvider<any>,
+      provider: OAuthProvider<TestEnv>,
       env: any,
       ctx: MockExecutionContext,
       code: string,
@@ -7014,7 +7014,7 @@ describe('OAuthProvider', () => {
     }
 
     async function refreshTokens(
-      provider: OAuthProvider<any>,
+      provider: OAuthProvider<TestEnv>,
       env: any,
       ctx: MockExecutionContext,
       refreshToken: string,
@@ -7042,7 +7042,7 @@ describe('OAuthProvider', () => {
     }
 
     async function callApi(
-      provider: OAuthProvider<any>,
+      provider: OAuthProvider<TestEnv>,
       env: any,
       ctx: MockExecutionContext,
       accessToken: string


mattzcarey:
gonna release this as a minor but the last one was massive so waiting a couple of days to test this better.

PR #158, opened 2026-02-27
fix: make CIMD explicitly opt-in via clientIdMetadataDocumentEnabled

### LINKED ISSUE #156: CIMD should fall back to DCR on failure, and be explicit opt-in
## Problem

Two issues with the current CIMD (Client ID Metadata Document) implementation:

### 1. CIMD failure is fatal — should fall back to DCR on client side. we just error out. 

When a client uses a URL as its `client_id` and CIMD fetching fails (e.g. the metadata document exceeds the 5KB size limit, or the fetch times out, or the URL returns an error), the library throws an error that propagates as a 500 to the user. 

We're seeing this in production where clients using URL-based `client_id`s (e.g. `https://claude.ai/oauth/mcp-oauth-client-metadata`) hit the 5KB response size limit and get a "Server Error" on `/authorize`. Maybe due to blot blocking?

### 2. No explicit way to disable CIMD

The only way to disable CIMD is to **remove the `global_fetch_strictly_public` compatibility flag** from `wrangler.jsonc`. This is a very indirect mechanism:

CIMD should be opt in. 

- The flag's purpose is SSRF protection, not CIMD control
- Removing it has side effects on all `fetch()` calls in the worker, not just CIMD
- There's no documentation that removing this flag is how you opt out of CIMD
- It's not discoverable — you have to read the library source to figure this out

There should be an explicit option on `OAuthProvider`, e.g.:

```typescript
new OAuthProvider({
  clientIdMetadataDocument: false, // or { enabled: false }
  // ...
})
```

mattzcarey:
## Summary

Fixes #156

Three changes to make CIMD (Client ID Metadata Document) robust and explicitly opt-in:

### 1. Explicit opt-in via `clientIdMetadataDocumentEnabled`

CIMD previously auto-enabled whenever the `global_fetch_strictly_public` compatibility flag was present. This caused 500 crashes for servers where URL-shaped `client_id`s (e.g. `https://claude.ai/oauth/mcp-oauth-client-metadata`) hit bot-protected endpoints, since there was no way to disable CIMD independently of the compat flag.

- Added `clientIdMetadataDocumentEnabled?: boolean` to `OAuthProviderOptions` (defaults to `false`)
- When **not enabled** (default): URL-formatted `client_id`s fall through to standard KV lookup — no fetch, no crash
- When **enabled without compat flag**: throws a clear error telling you to add `global_fetch_strictly_public`
- When **enabled with compat flag**: CIMD works as before
- Metadata endpoint reports `client_id_metadata_document_supported: true` only when **both** the option and compat flag are set

### 2. Graceful CIMD fetch failure handling

When CIMD is enabled and the metadata fetch fails (size limit exceeded, timeout, HTTP error, invalid JSON, network error, metadata validation failure), `getClient()` now catches the error and returns `null` instead of throwing. This means:

- Callers see "Invalid client" instead of an unhandled 500 error
- The token endpoint returns a proper `invalid_client` OAuth error response (HTTP 401), which the typescript-sdk client can catch and retry with DCR
- No more production crashes from bot-protected endpoints

### 3. Structured error logging

CIMD fetch failures are logged via `console.warn` with the client URL and error message, so operators can debug issues without the error propagating as a 500:

```
CIMD fetch failed for https://client.example.com/metadata.json: Failed to fetch client metadata: HTTP 404
```

### Usage

```typescript
new OAuthProvider({
  clientIdMetadataDocumentEnabled: true, // opt-in to CIMD
  // ...
})
```

### Source changes (`src/oauth-provider.ts`)

- Added `clientIdMetadataDocumentEnabled` option to `OAuthProviderOptions`
- `getClient()`: new CIMD opt-in gate → KV fallback when disabled; try/catch with `console.warn` when enabled
- Metadata endpoint: `client_id_metadata_document_supported` now requires both option AND compat flag

## Test plan

- [x] All 251 tests pass (5 new test groups added)
- [x] **Explicit opt-in tests**: default falls through to KV, metadata reports false, enabled+flag works, enabled-no-flag throws
- [x] **Error logging tests**: verifies `console.warn` is called with client URL and error details for HTTP failures, timeouts, size limits, and validation failures
- [x] **Token endpoint test**: CIMD failure returns proper `invalid_client` OAuth error (401)
- [x] Existing CIMD success tests updated to use `clientIdMetadataDocumentEnabled: true`
- [x] Existing failure tests updated to expect "Invalid client" instead of specific internal errors
- [x] Typecheck passes
- [x] Prettier passes
- [x] **Deployed to staging** (`staging.mcp.cloudflare.com`) and tested with Claude.ai as client — both CIMD disabled and enabled modes work correctly

### DIFF
--- a/.changeset/client-registration-rfc7591.md
+++ b/.changeset/client-registration-rfc7591.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Include `client_secret_expires_at` and `client_secret_issued_at` in dynamic client registration responses when a `client_secret` is issued, per RFC 7591 §3.2.1.
--- a/__tests__/oauth-provider.test.ts
+++ b/__tests__/oauth-provider.test.ts
@@ -551,6 +551,8 @@ describe('OAuthProvider', () => {
       const registeredClient = await response.json<any>();
       expect(registeredClient.client_id).toBeDefined();
       expect(registeredClient.client_secret).toBeDefined();
+      expect(registeredClient.client_secret_expires_at).toBe(0);
+      expect(registeredClient.client_secret_issued_at).toEqual(expect.any(Number));
       expect(registeredClient.redirect_uris).toEqual(['https://client.example.com/callback']);
       expect(registeredClient.client_name).toBe('Test Client');
 
@@ -583,6 +585,8 @@ describe('OAuthProvider', () => {
       const registeredClient = await response.json<any>();
       expect(registeredClient.client_id).toBeDefined();
       expect(registeredClient.client_secret).toBeUndefined(); // Public client should not have a secret
+      expect(registeredClient.client_secret_expires_at).toBeUndefined();
+      expect(registeredClient.client_secret_issued_at).toBeUndefined();
       expect(registeredClient.token_endpoint_auth_method).toBe('none');
 
       // Verify the client was saved to KV
--- a/src/oauth-provider.ts
+++ b/src/oauth-provider.ts
@@ -2715,9 +2715,11 @@ class OAuthProviderImpl<Env = Cloudflare.Env> {
       client_id_issued_at: clientInfo.registrationDate,
     };
 
-    // Only include client_secret for confidential clients
+    // Only include client_secret for confidential clients (RFC 7591 §3.2.1)
     if (clientSecret) {
       response.client_secret = clientSecret; // Return the original unhashed secret
+      response.client_secret_expires_at = 0; // Secret does not expire
+      response.client_secret_issued_at = clientInfo.registrationDate;
     }
 
     return new Response(JSON.stringify(response), {


mattzcarey:
## Tested on staging

Deployed to `staging.mcp.cloudflare.com` (cloudflare/mcp) and verified:

1. **CIMD disabled (default):** URL-shaped `client_id`s (e.g. `https://claude.ai/oauth/mcp-oauth-client-metadata`) no longer crash the server — they fall through to KV lookup gracefully.
2. **CIMD enabled (`clientIdMetadataDocumentEnabled: true`):** CIMD works as expected with the compat flag. Failures are logged via `console.warn` and return "Invalid client" instead of 500.
3. **Typecheck passes** with the new option in cloudflare/mcp.

Tested with Claude.ai as the client.

deathbyknowledge (approved):

PR #184, opened 2026-04-09
feat: add clientRegistrationCallback for customizing DCR

### LINKED ISSUE #162: Allow overriding `clientRegistrationEndpoint` with `defaultHandler`
We need to customize dynamic client registration to meet our requirements. It would be great if there was an option to override the default handler. Not sure what the best design here is, hence I didn't open a PR right away. Perhaps an additional config item: `customClientRegistrationEndpoint` or similar?

Talador12:
Adds `clientRegistrationCallback` to `OAuthProviderOptions` — same pattern as `tokenExchangeCallback`, applied to DCR.

- Callback invoked after validation, before KV storage
- Can reject with custom error code/status, or override safe metadata fields
- Overrides use a Pick allowlist (clientName, contacts, URI metadata, etc.) — security fields like redirectUris/clientSecret/tokenEndpointAuthMethod can't be set
- Override values are revalidated before storage, including http(s)-only URI validation for URI metadata fields
- Runtime enforcement mirrors the type constraint

Closes #162.

### DIFF
--- a/.changeset/dcr-callback.md
+++ b/.changeset/dcr-callback.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': minor
+---
+
+Add `clientRegistrationCallback` for validating or rejecting dynamic client registrations before storage. Follows the same pattern as `tokenExchangeCallback`. Closes #162.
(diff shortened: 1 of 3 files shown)


mattzcarey:
@Talador12 can you have a look at my latest commit. I found a couple of bits of inconsistency but want you to check you are happy or if they were intentional.

mattzcarey (approved):
LGTM. Rebased on current main and added validation for callback metadata overrides so URI fields keep the same http(s)-only constraints before storage.

PR #185, opened 2026-04-09
feat: add client credentials grant (RFC 6749 §4.4)

Talador12:
Adds `client_credentials` grant (RFC 6749 §4.4) for M2M auth.

- `allowClientCredentialsGrant: true` to enable
- Confidential clients only, no refresh token
- Synthetic userId uses `cc_${sha256(clientId)}` to avoid collisions with real users and the `:` token separator
- `tokenExchangeCallback` fires with `grantType: 'client_credentials'` for customization
- Grant TTL matches access token TTL


mattzcarey:
@Talador12 same here — pushed a follow-up commit (b8cc88a) with a couple of tweaks. Have a look and let me know if anything was intentional.

CyberFlameGO (approved):

mattzcarey:

PR #187, opened 2026-04-09
fix: advertise fragment response mode when implicit flow is enabled

Talador12:
`response_modes_supported` now includes `fragment` when `allowImplicitFlow` is true. Implicit flow uses fragment redirect per RFC 6749 §4.2.2.


mattzcarey (approved):
LGTM — RFC 8414 §2 is the metadata requirement here, with RFC 6749 §4.2.2 explaining the implicit-flow fragment response.

PR #189, opened 2026-04-09
feat: add token introspection endpoint (RFC 7662)

Talador12:
Adds a token introspection endpoint (RFC 7662), enabled with `introspectionEndpoint`.

## Summary

- Adds `introspectionEndpoint?: string` provider option
- Handles POSTed form-encoded introspection requests using existing token-endpoint client authentication parsing
- Requires a confidential client; public clients get `401 unauthorized_client`
- Returns `{ active: false }` for invalid, expired, unknown, or non-owned tokens
- Returns RFC 7662-style active token metadata for owned access tokens:
  - `active`
  - `iss`
  - `client_id`
  - `token_type`
  - `exp`
  - `iat`
  - `sub`
  - `scope`
  - `aud` when present
- Advertises `introspection_endpoint` and `introspection_endpoint_auth_methods_supported` in authorization-server metadata when configured
- Uses the repo's sensitive response cache headers (`Cache-Control: no-store`, `Pragma: no-cache`) for introspection responses

## Verification

- `npm run check` ✅ — 507 tests passing
- `npm run build` ✅
- `npx prettier --check src/oauth-provider.ts __tests__/oauth-provider.test.ts .changeset/token-introspection.md` ✅
- `git diff --check origin/main...HEAD` ✅

## Note

Do not merge without Matt's explicit approval.

### DIFF
--- a/.changeset/token-introspection.md
+++ b/.changeset/token-introspection.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': minor
+---
+
+Add token introspection endpoint (RFC 7662). Enable with `introspectionEndpoint`. Confidential clients only; returns `active: false` for tokens not owned by the requesting client.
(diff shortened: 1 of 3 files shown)


mattzcarey:

PR #190, opened 2026-04-09
fix: honor token_type_hint in revocation requests (RFC 7009 §2.1)

Talador12:
Revocation endpoint now uses `token_type_hint` to check the hinted type first (RFC 7009 §2.1). Incorrect hints fall through gracefully.


Talador12:
Merged into #188 — client ownership check and token_type_hint are now in the same PR since they both modify revokeToken.

Issue #191, opened 2026-04-09
feat: DPoP support (RFC 9449) — proof-of-possession token binding


Talador12:
## Summary
Would be interested in contributing DPoP (Demonstrating Proof of Possession) support per RFC 9449. This binds access tokens to a client's key pair, so leaked tokens can't be used without the corresponding private key.

## Motivation
- Directly addresses the security concern in #43 (refresh token rotation) — DPoP makes stolen tokens useless without the private key
- OAuth 2.1 best practice for public client security
- Relevant for MCP clients running in environments where token theft is a concern

## Proposed Design
1. New `allowDPoP: boolean` option (default `false`)
2. At the **token endpoint**: if client sends a `DPoP` header, verify the proof and store the JWK thumbprint (`jkt`) on the token in KV
3. At **API request validation**: if the token has a stored `jkt`, require and verify a `DPoP` proof header. Tokens without `jkt` remain bearer tokens (backward compatible).
4. Metadata: advertise `dpop_signing_alg_values_supported: ['RS256', 'ES256']`

## KV Schema Impact
Adds a `dpopJkt?: string` field to the Token record. This is backward compatible — existing tokens without `dpopJkt` continue working as bearer tokens.

## Notes
- This touches token storage (KV schema), token creation, and token validation — all "ask first" categories per AGENTS.md
- I have a working zero-dependency DPoP implementation (139 lines, pure Web Crypto) that can be adapted
- Happy to discuss the design before coding

References #43

Issue #43, opened 2025-06-06
Allowing the previous refresh token makes RT rotation useless


neilmadden-teya:
The point of requiring a new refresh token to be issued each time is to allow some (weak) level of protection to public clients using refresh tokens. OAuth 2.1 only requires this for non-confidential clients. As [per section 4.3.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-13#refresh-token-endpoint-extension):

> If a refresh token is compromised and subsequently used by both the attacker and the legitimate client, one of them will present an invalidated refresh token, which will inform the authorization server of the breach. The authorization server cannot determine which party submitted the invalid refresh token, but it will revoke the active refresh token as well as the access authorization grant associated with it. This stops the attack at the cost of forcing the legitimate client to obtain a fresh authorization grant.

Allowing the previous refresh token to be used entirely disables this protection.

Some OAuth providers have tried to solve this issue by putting a strict time limit on how long the previous RT remains valid after it has been rotated.

kentonv:
This decision is discussed in the readme:

https://github.com/cloudflare/workers-oauth-provider?tab=readme-ov-file#single-use-refresh-tokens

There is a tension here between risk management and availability.

The requirements in the spec are well-meaning but ignore the realities of distributed systems. The requirement will necessarily lead to grants being incorrectly revoked as a result of simple network partitions or machine failures -- situations where no attack took place. Pulling a network cable could cause someone's system to suddenly and permanently stop working (until they manually issue a new grant). I felt this was an unreasonable trade-off.

Note that this was my intentional decision, not something the LLM did. The LLM implemented the spec correctly and I asked for the change. It is perhaps arrogant of me to think I know better than the spec authors and perhaps I will end up losing this argument and looking bad but if so that's 100% on me the human, not the LLM.

kentonv:
> OAuth 2.1 only requires this for non-confidential clients.

That's not quite right: It requires either strict refresh token rotation, or that the implementation "cryptographically binds the refresh token to a certain client, e.g., by utilizing DPoP [RFC9449] or mTLS [RFC8705]."

[...]

PR #193, opened 2026-04-09
feat: add DPoP support (RFC 9449) for proof-of-possession token binding

Talador12:
Adds DPoP proof-of-possession (RFC 9449). Stolen tokens are useless without the client's private key.

- `allowDPoP: true` to enable, backward compatible (no DPoP header = bearer token)
- Token endpoint: verifies proof, stores JWK thumbprint on token and grant
- Resource server: accepts both `Authorization: Bearer` and `Authorization: DPoP` schemes, enforces proof + `ath` for bound tokens
- Refresh: grant-level key binding prevents switching keys (RFC 9449 §6)
- jti replay protection via KV with 5min TTL
- Rejects JWKs containing private key material
- RS256 and ES256, `dpop_signing_alg_values_supported` in metadata

References #43, #191.

### DIFF
--- a/.changeset/dpop-support.md
+++ b/.changeset/dpop-support.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': minor
+---
+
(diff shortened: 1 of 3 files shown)


mattzcarey:

PR #199, opened 2026-05-01
feat: convert OAuthError thrown from tokenExchangeCallback into structured /token responses

mattzcarey:
## Summary

Convert `OAuthError` thrown from `tokenExchangeCallback` into structured OAuth `/token` error responses instead of letting intentional callback failures bubble up as `500 Internal Server Error`.

There is one supported user-facing way to do this:

```ts
throw new OAuthError(code, {
  description,
  statusCode,
  headers,
});
```

This PR also converts transient token-storage KV write rate limits into structured retryable OAuth responses, so `/token` does not leak Worker 500s when KV returns `429 Too Many Requests` while issuing or rotating tokens.

## Why

When `tokenExchangeCallback` runs during `authorization_code`, `refresh_token`, or token-exchange grants and throws — for example because an upstream OAuth provider rejected a refresh token or rate-limited us — the error currently bubbles to the outer fetch handler and is served as a `500`.

Clients then have no machine-readable OAuth error telling them to stop retrying the same dead refresh token or to back off temporarily, which can create refresh-token retry storms against upstream providers.

The library already converted its internal `OAuthError` inside `handleTokenExchangeGrant`; this PR makes the behavior consistent at the token endpoint boundary and exposes `OAuthError` as the public callback API for intentional OAuth failures.

## API

```ts
import { OAuthError, OAuthProvider } from '@cloudflare/workers-oauth-provider';

new OAuthProvider({
  // …
  tokenExchangeCallback: async (options) => {
    if (options.grantType === 'refresh_token') {
      // refreshUpstream() may throw OAuthError from any depth.
      return { newProps: await refreshUpstream(options.props) };
    }
  },
});

async function refreshUpstream(props) {
  const res = await fetch(/* upstream token endpoint */);

  if (res.status === 401) {
    throw new OAuthError('invalid_grant', {
      description: 'upstream refresh token is invalid',
    });
  }

  if (res.status === 429) {
    throw new OAuthError('temporarily_unavailable', {
      description: 'upstream rate limited',
      statusCode: 429,
      headers: { 'Retry-After': res.headers.get('retry-after') ?? '60' },
    });
  }

  return await res.json();
}
```

### `OAuthError(code, options)`

- **`code`** — OAuth error code returned in the response `error` field.
  - This is intentionally a `string`, not runtime-validated. Applications can pass custom or extension error codes when their clients understand them.
  - For standard OAuth token-endpoint codes, this package exports `OAuthTokenErrorCode` as a helper type.
- **`options.description`** — human-readable text returned in `error_description` and used as the JavaScript `Error.message`.
- **`options.statusCode`** — HTTP status code. Defaults to `400`.
- **`options.headers`** — additional response headers.

`description` lives in the options object alongside `statusCode` and `headers`, rather than as a second positional argument, so the constructor stays clear as response metadata grows.

## Behavior

`handleTokenRequest` catches this package's exported `OAuthError` thrown anywhere below it and converts it into a structured OAuth `/token` response.

This applies to:

- `authorization_code`
- `refresh_token`
- token-exchange grant paths

Anything else thrown — plain `Error`, plain objects with a `code` field, app-local error classes, etc. — is re-thrown so unexpected failures stay visible as 500s. The provider does **not** catch-everything-and-return-400.

This keeps the parsing path deliberately simple: if users want a structured `/token` error, they import and throw `OAuthError` from `@cloudflare/workers-oauth-provider` rather than copying or re-implementing it.

### Token storage KV rate limits

KV `429 Too Many Requests` write failures while issuing tokens are treated as transient token endpoint failures and returned as:

```http
HTTP/1.1 429 Too Many Requests
Content-Type: application/json
Retry-After: 30
```

```json
{
  "error": "temporarily_unavailable",
  "error_description": "Token issuance is temporarily unavailable; retry shortly"
}
```

This currently covers grant saves and access-token writes used by token issuance/refresh paths.

## Why no returned `{ error }` form?

I prototyped a returned form and dropped it. As soon as an OAuth call is factored out of the callback body, every helper has to return and forward a discriminated union:

```ts
// Awkward
async function refreshUpstream(props): Promise<{ tokens: T } | { error: E }> {
  // …
}
const r = await refreshUpstream(props);
if ('error' in r) return { error: r.error }; // hand-pumping
```

Throwing `OAuthError` avoids that plumbing. Auth failure is genuinely exceptional from the happy path's perspective; this is exceptions for exceptional conditions, not exceptions for ordinary branching.

## Why no first-class `retryAfter` field or implicit default?

`Retry-After` remains just a header in `options.headers`.

Reasons:

1. **Two ways to set one header is a footgun** — `retryAfter: 12` plus `headers['Retry-After'] = '5'` would need precedence rules.
2. **It would be arbitrary special-casing** — why elevate `Retry-After` and not `Cache-Control`, `WWW-Authenticate`, etc.?
3. **`Retry-After` can be seconds or an HTTP-date** (RFC 7231 §7.1.3). Headers naturally support both.
4. **Call sites usually mirror upstream's header**, which is already a string.

There is no library-level default for caller-thrown `OAuthError`. If callers want retry guidance, they set the header explicitly. The only built-in `Retry-After: 30` added by this PR is for the provider's own transient KV write-rate-limit handling.

### New exports

- **`OAuthError`** — public error class for intentional OAuth token endpoint failures.
- **`OAuthErrorOptions`** — `{ description, statusCode?, headers? }`.
- **`OAuthTokenErrorCode`** — helper string-literal union of standard OAuth token-endpoint codes. This is exported for callers who want standard-code typing, but `OAuthError` itself accepts any string code.

### Versioning

`minor` bump via changeset: this widens the public API surface by exporting `OAuthError`, `OAuthErrorOptions`, and `OAuthTokenErrorCode`, and it changes transient token-storage KV write rate limits from Worker 500s into structured retryable OAuth responses.

The previously-internal `OAuthError` constructor was `(code, message)`. The new public constructor is `(code, options)`, with `options.description` used for both OAuth `error_description` and `Error.message`.

## Test plan

Verified locally:

- `npm run typecheck`
- `npm test -- __tests__/oauth-provider.test.ts` — 352 passing
- `npx prettier --check README.md __tests__/oauth-provider.test.ts src/oauth-provider.ts`

New/updated tests cover:

- Successful refresh exchanges still work end-to-end.
- Throwing `new OAuthError('invalid_grant', { description })` on `refresh_token` returns structured 400.
- `headers: { 'Retry-After': '12' }` on a 429 is preserved.
- No `Retry-After` is added by default on caller-thrown 429 `OAuthError`.
- Custom headers pass through.
- Same structured error path for `authorization_code`.
- Same structured error path for token-exchange grant.
- `OAuthError` thrown deep in helper call stacks propagates correctly.
- Custom string codes passed to the exported `OAuthError` are preserved in structured responses.
- App-local OAuthError classes are not special-cased; use the exported class.
- Plain `Error` is re-thrown / surfaces as 500.
- `Error` instances with `code`, plain objects with `code`, and other non-`OAuthError` shaped throws are re-thrown / surface as 500.
- KV 429 write failures during token issuance return structured `temporarily_unavailable` with `Retry-After: 30` instead of leaking Worker 500s.

This PR replaces the patch shipped in [cloudflare/mcp#112](https://github.com/cloudflare/mcp/pull/112), which can now use the new pkg.pr.new build.


threepointone (approved):
This looks good to me. Clanker said this, they seem like nits to me, so I'll leave it up to you whether you want to handle them (I kinda wouldn't). Ship it. 

Image

Issue #196, opened 2026-04-15
Few issues regarding token endpoint client authentication


ylemkimon:
* Token endpoint should respond with `invalid_request` if a parameter is included more than once. Currently, duplicate parameters become an array and are coerced into a comma-delimited string, which will fail validation later, but it should not rely on this (`body` is typed `any`, so Typescript does not catch this).
> ## RFC 6749
> **3.2 Token Endpoint**
> Request and response parameters MUST NOT be included more than once.

> **5.2. Error Response**
> invalid_request: The request ... repeats a parameter

https://github.com/cloudflare/workers-oauth-provider/blob/251d6412e746c7abcfcac662112e0d08a9976f7c/src/oauth-provider.ts#L1529-L1533

* It should respond with `invalid_request` if multiple credentials are included.

> **2.3. Client Authentication**
> The client MUST NOT use more than one authentication method in each request.

> **5.2. Error Response**
> invalid_request: The request ... includes multiple credentials

https://github.com/cloudflare/workers-oauth-provider/blob/251d6412e746c7abcfcac662112e0d08a9976f7c/src/oauth-provider.ts#L1541-L1551

* In `String#split()`, any leftover text after the limit is not included in the array at all. However, in Basic auth, the password can contain colons. (This does not affect client secrets generated by this library)

https://github.com/cloudflare/workers-oauth-provider/blob/251d6412e746c7abcfcac662112e0d08a9976f7c/src/oauth-provider.ts#L1544

* Client credentials use `application/x-www-form-urlencoded` encoding, which encodes a space into `+` not `%20`, so they should be replaced manually. (This does not affect client secrets generated by this library)

> **2.3.1. Client Password**
> The client identifier is encoded using the "application/x-www-form-urlencoded" encoding algorithm per [Appendix B](https://datatracker.ietf.org/doc/html/rfc6749#appendix-B), and the encoded value is used as the username; the client password is encoded using the same algorithm and used as the password.

https://github.com/cloudflare/workers-oauth-provider/blob/251d6412e746c7abcfcac662112e0d08a9976f7c/src/oauth-provider.ts#L1545-L1546

mattzcarey:
closing as fixed :)

PR #207, opened 2026-05-15
fix: validate token endpoint client authentication

EfeDurmaz16:
## Summary

Fixes part of #196 by tightening token endpoint request parsing for RFC 6749 client authentication rules.

- reject repeated token endpoint parameters as `invalid_request`, while preserving repeated `resource` support from RFC 8707
- reject requests that combine HTTP Basic auth with form client credentials
- decode HTTP Basic client credentials using application/x-www-form-urlencoded semantics and preserve colons in the password side

## Verification

- `npx vitest run __tests__/oauth-provider.test.ts --testNamePattern "Authorization Code Flow Exchange"`
- `npm run check`
- `npm run build`

### DIFF
--- a/.changeset/fair-donuts-drum.md
+++ b/.changeset/fair-donuts-drum.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Tighten token endpoint client authentication parsing for RFC 6749 compliance.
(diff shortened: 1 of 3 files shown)


mattzcarey (approved):
LGTM — tightens token endpoint client authentication parsing in line with RFC 6749/RFC 8707, with coverage for duplicate parameters, mixed auth methods, and Basic credential decoding.

PR #203, opened 2026-05-12
Experiment: support MCP enterprise-managed authorization

irvinebroque:
## Summary
- Experimental implementation of MCP Enterprise-Managed Authorization in `workers-oauth-provider` using the JWT bearer grant and ID-JAG validation.
- Adds trusted enterprise issuer configuration, WebCrypto JWT/JWKS verification for RS256 and ES256, claim mapping, resource/client/audience validation, scope validation, and KV-backed `jti` replay markers.
- Adds README docs, a minor changeset, an implementation plan, and comprehensive token endpoint coverage.

## Intent
This draft PR is **not intended to be merged**. It is an experiment to understand what changes would be required in `workers-oauth-provider` and downstream Cloudflare Agents SDK integrations.

## Validation
- `npm run check`
- `npm run build`
- `git diff --check`

## Notes / Known Caveats
- The MCP extension is still draft, so the API is explicitly experimental and opt-in.
- Replay protection uses KV read/write markers; KV does not provide an atomic compare-and-set primitive, so production-grade single-use `jti` enforcement may need a stronger storage primitive if strict concurrency guarantees are required.
- The current opaque token format uses `:` separators, so the enterprise mapper rejects mapped user IDs containing `:`.
- This PR only implements the provider-side pieces. Cloudflare Agents SDK still needs MCP capability declaration helpers/docs, and client-side ID-JAG acquisition would be separate work.

PR #208, opened 2026-05-15
feat(ema): restructure enterprise-managed auth into Result-typed modules

mattzcarey:
## Summary

Adds support for MCP [Enterprise-Managed Authorization](https://modelcontextprotocol.io/extensions/auth/enterprise-managed-authorization) (EMA / ID-JAG) via the RFC 7523 `urn:ietf:params:oauth:grant-type:jwt-bearer` token endpoint grant. Purely additive — no existing API or behavior changes. EMA only activates when the new `enterpriseManagedAuthorization` option is configured.

Builds on the protocol work in #203 (closed), restructured into focused modules under `src/ema/` with a `Result`-typed pipeline of pure validators. Diff vs `main`: +3091 / -11.

### New token-endpoint grant
At `POST /token`, accepts `grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer` with `assertion=<ID-JAG>`. The flow:

1. Parse compact JWT, enforce 16 KiB cap (`parser.ts`)
2. Validate JOSE header — `typ=oauth-id-jag+jwt`, allow-listed `alg` (`validators.validateIdJagHeader`)
3. Resolve `iss` through the deployer-supplied `trustedIssuers` resolver (`validators.resolveTrustedIssuer`)
4. Fetch + cache the issuer's JWKS, select the matching JWK, verify the signature with WebCrypto (`jwks.ts`, `signature.ts`)
5. Validate every required claim: `iss / sub / aud / resource / client_id / jti / exp / iat / nbf? / scope?` (`validators.validateIdJagClaims`)
6. Record the `jti` in the KV-backed replay store (`jti.ts`)
7. Invoke `mapClaims` to translate IdP identity → AS user + props + scope (`validators.validateEmaMapperResult`)
8. Mint and return an access token; refresh tokens are intentionally not issued (the ID-JAG itself is the renewable grant)

### New `src/ema/` modules

| File | Purpose |
|------|---------|
| `constants.ts` | All `EMA_*` constants co-located, including `EMA_ID_JAG_GRANT_PROFILE` |
| `result.ts` | `Result<T,E>` + 27-variant `EmaValidationError` + `emaErrorToWire` |
| `types.ts` | Public + internal types — `EmaOptions`, `EmaTrustedIssuer`, `EmaClaimsMapper`, etc. |
| `parser.ts` | `parseIdJag` (pure, Result-returning) |
| `validators.ts` | Pure validators: header / issuer resolve / claims / scope / mapper / TTL |
| `signature.ts` | `selectJwk` + `verifyIdJagSignature` (WebCrypto) |
| `jwks.ts` | `EmaJwksProvider` interface + default fetcher with anti-DoS cool-down |
| `jti.ts` | `EmaJtiStore` interface + default KV-backed impl |
| `util.ts` | `sha256Hex` |

The handler is a ~120-line `runEmaPipeline` orchestrator chaining the validators with explicit `if (!r.ok) return r;`. No library, no hidden control flow.

### AS metadata advertising

`/.well-known/oauth-authorization-server` now advertises (when EMA is configured):
- `urn:ietf:params:oauth:grant-type:jwt-bearer` in `grant_types_supported`
- `urn:ietf:params:oauth:grant-profile:id-jag` in `authorization_grant_profiles_supported`

### Public API surface

```ts
new OAuthProvider({
  // ... existing options ...
  enterpriseManagedAuthorization: {
    trustedIssuers: async ({ iss, env, request, clientInfo }) => /* EmaTrustedIssuer | null */,
    mapClaims: async ({ claims, env, request, clientInfo }) => ({
      userId: 'user-123',
      scope: ['read', 'write'],
      props: { /* per-request props */ },
      accessTokenTTL: 86_400, // optional, falls back to AS default
    }),
    jwksCacheTtlSeconds: 300,         // optional
    clockSkewSeconds: 60,             // optional
    maxAssertionLifetimeSeconds: 300, // optional
  },
});
```

Both callbacks are `Promise`-returning — locked to async so KV/D1/IdP federation lookups don't require a later API expansion. For a fixed list, write a one-line closure with `async ({ iss }) => list.find(...) ?? null`.

The existing top-level `onError` hook on `OAuthProviderOptions` fires for EMA failures too — the typed `EmaValidationError` reason flows via the new (backwards-compatible) `error.internal: { category, reason, detail? }` field, which the wire response never sees.

### Security properties

- **Confused-deputy guard** on the resolver — `resolved.issuer === iss` is enforced
- **Per-request structural validation** of resolver output — HTTPS `issuer` AND `jwksUri` (RFC 8414 §2), supported algorithm
- **JWKS force-refresh cool-down** — per-issuer 30s default rate-limit. Defends against an attacker sending many assertions with random `kid` values to amplify load on the IdP
- **`nbf` claim support** per RFC 7523 §3 rule 5
- **`exp`, `iat`, `nbf` all honor `clockSkewSeconds`** (RFC 7523 §3 rule 4)
- **`alg=none` rejected**; algorithm allowlist enforced (`RS256`, `ES256`); JWK `kty` must match the header `alg`
- **Resource pinning** at construction time — EMA requires `resourceMetadata.resource` to be declared, preventing an EMA-configured provider from accepting any RFC 8707-shaped `resource` claim
- **`mapClaims` exceptions caught** — translated to `invalid_grant: Assertion was not authorized`, not 500
- **`Cache-Control: no-store` + `Pragma: no-cache`** on every EMA token response (RFC 6749 §5.1)
- **Wire-level errors collapse to generic `invalid_grant: Invalid assertion`** (RFC 6749 §5.2) — the rich tagged reason flows server-side to the deployer's `onError.internal` sink for debugging/alerting

### Why a `Result` type instead of throws

Throwing across 18 sites with `OAuthError('invalid_grant', 'Invalid assertion')` makes the validation path harder to audit. The Result pipeline keeps the same wire-level behavior (generic error) but makes each check inspectable in isolation. Two helpers (`ok` / `err`), zero new dependencies.

## Tests

| | |
|---|---|
| Total | **435 passing** |
| Integration (`__tests__/oauth-provider.test.ts`) | 379 — includes new EMA construction-time checks, metadata advertising, end-to-end ID-JAG exchange, `mapper_threw`, and `onError.internal` smoke test |
| EMA unit (`__tests__/ema-validators.test.ts`) | 56 — each pure validator pinned without spinning up `OAuthProvider`, mocking fetch, or seeding KV |

## Compatibility

- **No breaking changes** — every existing API is unchanged
- Feature is **opt-in** via the new `enterpriseManagedAuthorization` option
- New AS metadata fields appear only when EMA is configured
- Followup: [#211](https://github.com/cloudflare/workers-oauth-provider/issues/211) tracks rolling the new `onError.internal` shape out to other generic-error paths

## Test plan

- [x] `npm run check` — 435/435 tests pass
- [ ] Review the `trustedIssuers` resolver-only API
- [ ] Review the anti-DoS JWKS cool-down (30s default)
- [ ] End-to-end exchange against a real IdP

### DIFF
--- a/.changeset/enterprise-managed-authorization.md
+++ b/.changeset/enterprise-managed-authorization.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': minor
+---
+
+Experimentally support MCP Enterprise-Managed Authorization ID-JAG assertions through the JWT bearer grant.
--- a/README.md
+++ b/README.md
@@ -388,7 +388,7 @@ Setup:
 2. Set `resourceMetadata.resource` to the MCP endpoint URL (required when EMA is enabled).
 3. Implement `trustedIssuers` as a resolver — for multi-tenant deployments it can read `env` / `clientInfo` to look up per-tenant IdP config without redeploying.
 
-The AS enforces `resolved.issuer === iss` (confused-deputy guard) and validates ID-JAG `typ`, signature, audience, client binding, resource, `exp` / `iat` / `nbf`, max lifetime, and `jti` replay. Access token TTL is clamped to the assertion lifetime; refresh tokens are not issued for this grant. Default in-memory JWKS cache and KV-backed JTI store can be replaced via `jwksProvider` / `jtiStore` adapters.
+The AS enforces `resolved.issuer === iss` (confused-deputy guard) and validates ID-JAG `typ`, signature, audience, client binding, resource, `exp` / `iat` / `nbf`, max lifetime, and `jti` replay. Refresh tokens are not issued for this grant — the ID-JAG itself is the renewable assertion.
 
 Experimental — the MCP extension is still a draft.
 
--- a/README.md
+++ b/README.md
@@ -365,7 +365,7 @@ new OAuthProvider({
   // ... other options ...
   resourceMetadata: { resource: 'https://mcp.example.com/mcp' },
   enterpriseManagedAuthorization: {
-    trustedIssuers: ({ iss }) =>
+    trustedIssuers: async ({ iss }) =>
       iss === 'https://idp.example.com'
         ? { issuer: iss, jwksUri: 'https://idp.example.com/.well-known/jwks.json', algorithms: ['RS256'] }
         : null,
--- a/README.md
+++ b/README.md
@@ -117,6 +117,10 @@ export default new OAuthProvider({
   // Defaults to false.
   allowTokenExchangeGrant: false,
 
+  // Optional: Experimental MCP Enterprise-Managed Authorization support.
+  // When enabled, the token endpoint accepts ID-JAG JWTs with the JWT bearer grant.
+  enterpriseManagedAuthorization: undefined,
+
   // Optional: Explicitly enable Client ID Metadata Document (CIMD) support.
   // When true, URL-formatted client_ids will be fetched as metadata documents.
   // Requires the 'global_fetch_strictly_public' compatibility flag.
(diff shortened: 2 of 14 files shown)


ciamshrek:
Hey @mattzcarey, it might be more desirable to allow 

```
    trustedIssuers: [
      {
        issuer: 'https://acme.idp.example',
        jwksUri: 'https://acme.idp.example/.well-known/jwks.json',
        algorithms: ['RS256', 'ES256'],   // optional, defaults to ['RS256']
        audience: 'https://auth.chat.example/',  // optional, defaults to AS issuer
      },
    ],
```

to be a provider / function or instead allow this to be introspected from the incoming token and then validated against an existing list. This would allow users that have B2B Identity Providers (eg tenant A -> idPA or User A -> idP B)  without having to modify the static config.

mattzcarey:
@ciamshrek thanks for the review.

I've pushed [`ba01920`](https://github.com/cloudflare/workers-oauth-provider/pull/208/commits/ba01920) — `trustedIssuers` is now always a resolver function. The static array shape is gone; there's no union or fallback. One shape, one way of writing it.

```ts
trustedIssuers: ({ iss, env, request, clientInfo }) => EmaTrustedIssuer | null | Promise<…>
```

For a fixed list it's a one-line closure:

```ts
const idps = [{ issuer: 'https://idp.example.com', jwksUri: '...' }];
trustedIssuers: ({ iss }) => idps.find((i) => i.issuer === iss) ?? null,
```

mattzcarey:
/bonk review pls — addressed all 11 findings in d571d0d (3 security/correctness + cleanup + docs + tests + the 3 non-blocking API surface items)

mattzcarey:
/bonk review pls — `febbad3` locks `trustedIssuers` and `mapClaims` to async-only (matching the resolver shape). Single canonical signature, drops the `T | Promise<T>` union before release.

mattzcarey:
/bonk review pls — `ea86642` tightens the provider model (drops adapter result types, force_refresh_throttled, translation helpers; providers speak Result directly) and inlines parseScopeGrammar. `7e5e637` addressed your 7 previous findings.

PR #216, opened 2026-06-01
feat: add coalesceRefreshTokenExchange for non-idempotent upstream callbacks

mattzcarey:
## Problem

When a Worker's `tokenExchangeCallback` redeems a **single-use, rotating** upstream refresh token (i.e. the Worker is itself a client of another OAuth server — like the Cloudflare MCP server), the `refresh_token` grant becomes non-idempotent. Two failure modes produce a constant stream of `invalid_grant` errors:

1. **Shared tokens / concurrent refresh.** Two sessions (e.g. two chat windows) share one downstream refresh token and refresh at the same time. Both decrypt the same upstream credentials and both invoke the callback, racing to redeem the same single-use upstream token. One wins; the other gets `invalid_grant` from upstream.
2. **Lost responses / retries.** A client never receives a refresh response and retries with its (now *previous*) refresh token. Today the callback runs again and redeems the already-rotated upstream token, cascading further rotations — a self-sustaining stream of errors.

## Fix

Add an opt-in `coalesceRefreshTokenExchange` provider option (default `false`, so existing behaviour is unchanged). When `true`:

- **Single-flight** — concurrent `refresh_token` requests presenting the same refresh token within a single isolate are coalesced: the callback runs once and every caller receives a clone of the same token response.
- **Idempotent replay** — a refresh presented with the grant's *previous* refresh token is treated as a retry of the rotation that already happened. The callback is skipped and a fresh access token is minted from the grant's current props, without rotating the refresh token or re-touching upstream.

## Honest limitation

KV is eventually-consistent with no compare-and-swap, so simultaneous refreshes landing on *different* isolates before either has persisted cannot be fully serialized. This change collapses the common same-isolate burst and makes client retries safe and upstream-free — it does not require a Durable Object.

## Changes

- `coalesceRefreshTokenExchange?: boolean` option, fully documented in the interface.
- `handleRefreshTokenGrant` split into a coalescing wrapper + `executeRefreshTokenGrant`; in-isolate single-flight via a `Map<key, Promise<Response>>` keyed by refresh token + client + scope/resource, with `Response.clone()` per caller.
- Idempotent-replay branch: skips the callback and rotation, returns the presented token with a fresh access token.
- 4 new tests (replay skips callback; replay doesn't rotate the grant; concurrent single-flight runs callback once; flag-off retry re-invokes callback). **357/357 tests pass, typecheck clean.**
- README section + changeset (minor).

Related: surfaced from recurring `Token refresh failed: 400 invalid_grant` in the Cloudflare MCP server, which is an OAuth client to Cloudflare upstream.

### DIFF
--- a/.changeset/coalesce-refresh-token-exchange.md
+++ b/.changeset/coalesce-refresh-token-exchange.md
@@ -1,0 +1,35 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Add `coalesceRefreshTokenExchange` option to harden providers whose
+`tokenExchangeCallback` performs a **non-idempotent** upstream operation during
+the `refresh_token` grant — most commonly redeeming a _single-use, rotating_
+upstream refresh token (i.e. the Worker is itself a client of another OAuth
+server, like the Cloudflare MCP server).
+
+Without this option, two concurrent refreshes that share the same downstream
+refresh token both decrypt the same upstream credentials and both invoke
+`tokenExchangeCallback`, racing to redeem the same single-use upstream token.
+One wins; the other gets `invalid_grant` from upstream. The losing client then
+retries with its (now _previous_) refresh token, which re-runs the callback
+against the freshly-rotated upstream token, redeeming it again and cascading
+further rotations — a self-sustaining stream of `invalid_grant` errors.
+
+When `coalesceRefreshTokenExchange: true`:
+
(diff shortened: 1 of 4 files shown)


mattzcarey:
should the replay have a ttl?

mattzcarey:
closing this. not gonna ship it rn

PR #218, opened 2026-06-03
Validate client metadata URI schemes during registration

mattzcarey:
## What

Validate the URI scheme of client metadata fields during client registration.

The `client_uri`, `logo_uri`, `policy_uri`, `tos_uri`, and `jwks_uri` fields were previously only checked to be strings. They are now required to be absolute `http:` or `https:` URLs, consistent with how `redirect_uris` are already validated via `validateRedirectUriScheme`.

This applies to both ingestion paths:
- Dynamic Client Registration (`/register`)
- Client ID Metadata Document (CIMD) fetching

Values using any other scheme (or non-absolute/malformed URLs) are now rejected with an `invalid_client_metadata` error.

## Why

These metadata fields are commonly surfaced in consent UIs as link or image targets. Restricting them to standard web URLs keeps non-`http(s)` schemes from flowing through the library to consumers, and brings their validation in line with the existing `redirect_uris` handling.

## How

- New `validateOptionalUriField` helper parses the value with `URL` and enforces an `http:`/`https:` protocol, returning `undefined` for omitted fields.
- Applied to all five metadata URI fields in both the registration and CIMD code paths.

## Tests

- Added a registration test asserting valid `http(s)` metadata URIs are accepted and persisted.
- Added a parameterized suite asserting each metadata URI field rejects unsafe/invalid values (`javascript:`, `data:`, `vbscript:`, `file:`, relative paths, non-URLs) with `invalid_client_metadata`.
- Full suite passes (`npm run check`): 471 tests.

A changeset is included (patch).

### DIFF
--- a/.changeset/validate-client-metadata-uri-schemes.md
+++ b/.changeset/validate-client-metadata-uri-schemes.md
@@ -1,0 +1,16 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Validate the URI scheme of client metadata fields during client registration.
+
+The `client_uri`, `logo_uri`, `policy_uri`, `tos_uri`, and `jwks_uri` fields
+were previously only checked to be strings. They are now required to be
+absolute `http:` or `https:` URLs, consistent with how `redirect_uris` are
+already validated. Registration (and Client ID Metadata Document ingestion)
+now rejects values using other schemes with an `invalid_client_metadata`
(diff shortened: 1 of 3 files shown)


threepointone (approved):

mattzcarey:
/bonk review this to check I didnt do anything silly.

PR #221, opened 2026-06-03
Preserve RFC 7591 §2.2 internationalized client metadata variants

### LINKED ISSUE #219: RFC 7591 §2.2 internationalized metadata variants (client_uri#ja, etc.) are silently dropped
## Summary

RFC 7591 §2.2 ("Human-Readable Client Metadata") allows internationalized variants of human-readable / URL metadata fields using a language-tag suffix, e.g.:

- `client_name#ja`
- `client_uri#ja-Jpan-JP`
- `tos_uri#fr`
- `policy_uri#de`

The library currently only reads the canonical, un-suffixed keys (`client_name`, `client_uri`, `logo_uri`, `tos_uri`, `policy_uri`, `jwks_uri`). The `#`-suffixed internationalized variants are silently ignored — both at registration ingest and when echoing metadata back.

## Impact

This is not a correctness/security problem today:

- Registrations that include i18n variants are accepted; the extra keys are just dropped.
- Validation logic (including any URI-scheme checks) only sees the canonical keys, so it won't accidentally reject an otherwise-valid i18n'd registration.

But it does mean the library is not fully RFC 7591 §2.2 compliant: a client that registers, say, a localized `client_uri#ja` and `client_name#ja` will not have those values stored or returned, so an authorization server built on top of this library can't present localized client info on the consent screen.

## Possible direction (if we want to support it)

- Preserve `#<lang-tag>`-suffixed variants for the human-readable fields during registration ingest, store them, and echo them back in the registration response.
- Apply the same per-field validation (e.g. URL-scheme checks for `*_uri#*`) to the localized variants as to the canonical key.

Filing for visibility / discussion; no urgency.

### References
- RFC 7591 §2.2 Human-Readable Client Metadata: https://www.rfc-editor.org/rfc/rfc7591#section-2.2

mattzcarey:
Closes #219.

## What

Preserve RFC 7591 §2.2 internationalized client metadata variants, which were previously silently dropped during registration.

Localized variants are expressed by appending a `#<BCP 47 language tag>` suffix to a metadata member name (e.g. `client_name#ja`, `tos_uri#fr`). This PR captures them for the five human-readable fields the RFC names — `client_name`, `client_uri`, `logo_uri`, `tos_uri`, `policy_uri` — and:

- stores them on the client record under a new optional `ClientInfo.i18n` map, keyed by the raw `field#tag` name (flat representation, mirrors the wire format);
- echoes them back as top-level `field#tag` members in the registration `201` response, alongside their canonical fields;
- handles them on both ingestion paths: Dynamic Client Registration (`/register`) and Client ID Metadata Document (CIMD) fetching.

## Validation

Each localized value is validated with the same rules as its canonical field:
- URI variants (`client_uri#*`, `logo_uri#*`, `tos_uri#*`, `policy_uri#*`) must be absolute `http:`/`https:` URLs (reuses `validateOptionalUriField` from #218).
- All variants must be strings.
- Language-tag case is preserved (tags are case-insensitive per BCP 47, so consumers should compare accordingly).
- Fields **not** in §2.2 (e.g. `jwks_uri`, `redirect_uris`, `contacts`) are not collected.

## Scope notes

- `jwks_uri` and `redirect_uris` are deliberately excluded — §2.2 covers only human-readable fields.
- `updateClient` replaces the whole `i18n` map when provided (RFC 7592 update semantics = replace), via the existing spread.

## Tests

- DCR: round-trip persist + response echo, BCP 47 tag-case preservation, omission when no variants present, non-§2.2 fields ignored, unsafe-scheme rejection on localized URI variants, non-string rejection.
- CIMD: accepts a document with localized variants; rejects a localized URI variant using an unsafe scheme.
- Full suite: `npm run check` → typecheck clean + 479 tests pass (+8 new).
- Patch changeset included.
- Also folds in @ask-bonk's nit from #218: passes the field name to the `client_name` `validateStringField` call in the DCR path.

### DIFF
--- a/.changeset/i18n-client-metadata-variants.md
+++ b/.changeset/i18n-client-metadata-variants.md
@@ -1,0 +1,19 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Preserve RFC 7591 §2.2 internationalized client metadata variants.
+
+Localized variants of the human-readable client metadata fields — expressed
+with a `#<BCP 47 language tag>` suffix on the member name (e.g.
+`client_name#ja`, `tos_uri#fr`) — were previously dropped during client
+registration. They are now captured for `client_name`, `client_uri`,
+`logo_uri`, `tos_uri`, and `policy_uri`, stored on the client record under a
+new optional `i18n` map (keyed by the raw `field#tag` name), and echoed back in
+the registration response alongside their canonical fields. The same handling
+applies to Client ID Metadata Document ingestion.
+
+Localized values are validated with the same rules as their canonical field:
+URI variants must be absolute `http:` or `https:` URLs, and all variants must
+be strings. Fields that are not part of RFC 7591 §2.2 (such as `jwks_uri` and
+`redirect_uris`) are not collected.
--- a/__tests__/oauth-provider.test.ts
+++ b/__tests__/oauth-provider.test.ts
@@ -813,6 +813,166 @@ describe('OAuthProvider', () => {
         });
       });
     });
+
+    describe('RFC 7591 §2.2 internationalized metadata variants', () => {
+      it('persists localized variants and echoes them in the response', async () => {
+        const clientData = {
+          redirect_uris: ['https://client.example.com/callback'],
+          client_name: 'Test Client',
+          'client_name#ja': 'テストクライアント',
+          'client_name#fr': 'Client de test',
+          client_uri: 'https://client.example.com',
+          'client_uri#ja': 'https://client.example.com/ja',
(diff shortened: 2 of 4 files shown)


mattzcarey:
/bonk review this to check I didnt do anything silly.

mattzcarey:
Thanks @ask-bonk — applied the defensive spread suggestion in 35426f2: the i18n variants are now copied in explicitly, skipping any key already present on the response, so a localized variant can't shadow a canonical member even if a future canonical field name were to contain a `#`. Left the second observation as-is since it was a positive callout.

PR #186, opened 2026-04-09
fix: add Cache-Control: no-store to token and error responses (RFC 6749 §5.1)

Talador12:
Adds `Cache-Control: no-store` to token responses, DCR responses, and error responses per RFC 6749 §5.1. Prevents intermediary caching of tokens and credentials.

PR #224, opened 2026-06-04
fix: prevent caching of token, credential, and error responses (RFC 6749 §5.1/§5.2)

mattzcarey:
## Summary

Supersedes #186, rebuilt on current `main`.

RFC 6749 §5.1 requires `Cache-Control: no-store` (and `Pragma: no-cache`) on **any response containing tokens, credentials, or other sensitive information**, so intermediaries don't cache them. The §5.2 error-response and RFC 7591 DCR examples show the same headers. This applies them to the provider-emitted OAuth responses.

## Scope

Headers added to:

- token endpoint **success** responses — authorization-code, refresh-token, RFC 8693 token-exchange (RFC 6749 §5.1 MUST)
- enterprise-managed authorization (EMA) JWT-bearer **success** response (EMA draft §5.2; shares the single `NO_CACHE_HEADERS` constant)
- dynamic client registration responses (carry `client_secret`)
- **all OAuth error responses** — applied centrally in `createErrorResponse`. RFC 6749 §5.2 has no normative MUST/SHOULD for this, but its error example (and OAuth 2.1 §3.2.4) shows `Cache-Control: no-store`; adding it is harmless defense-in-depth so OAuth state (`error`/`error_description`) isn't cached by any intermediary.

Deliberately **out of scope**:

- metadata discovery / protected-resource metadata (public, normally cacheable)
- the `completeAuthorization()` API and the authorization redirect — left unchanged

## Notes

- Cloudflare's own CDN doesn't cache JSON by default and only caches a narrow set of status codes (200/206/301, 302/303, 404/410), so this is primarily protection against other intermediaries/CDNs.
- `createErrorResponse` merges `{ ...NO_CACHE_HEADERS, ...callerHeaders }`, so caller-supplied headers (e.g. `Retry-After`, `WWW-Authenticate`) still take precedence.

## Implementation

- Single shared `NO_CACHE_HEADERS = { 'Cache-Control': 'no-store', Pragma: 'no-cache' }` constant.
- No public API changes.

## Tests

- New `expectNoCacheHeaders()` assertion helper.
- Coverage for DCR, authorization-code, refresh-token, token-exchange, and error responses.

Verification:

- `npm run typecheck`
- `npm run check` (typecheck + full vitest, 485 tests)
- `npx prettier --check` on touched files

### DIFF
--- a/.changeset/cache-control-no-store.md
+++ b/.changeset/cache-control-no-store.md
@@ -2,4 +2,4 @@
 '@cloudflare/workers-oauth-provider': patch
 ---
 
-Add `Cache-Control: no-store` and `Pragma: no-cache` to OAuth responses that contain tokens or credentials, per RFC 6749 §5.1 (token endpoint responses and dynamic client registration responses carrying `client_secret`).
+Add `Cache-Control: no-store` and `Pragma: no-cache` to OAuth responses that carry tokens, credentials, or OAuth state, matching the response examples in RFC 6749 §5.1/§5.2: token endpoint responses (success and error), dynamic client registration responses carrying `client_secret`, and EMA JWT-bearer token responses.
--- a/.changeset/cache-control-no-store.md
+++ b/.changeset/cache-control-no-store.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Add `Cache-Control: no-store` and `Pragma: no-cache` to OAuth responses carrying tokens, credentials, or other sensitive authorization data.
--- a/.changeset/cache-control-no-store.md
+++ b/.changeset/cache-control-no-store.md
@@ -2,4 +2,4 @@
 '@cloudflare/workers-oauth-provider': patch
 ---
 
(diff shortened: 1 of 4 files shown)


mattzcarey:
Thanks @ask-bonk — addressed:

1. **Stale description** — already rewritten before this review landed; the body now reflects what shipped (no `completeAuthorization`/README changes). That redirect-headers approach was reverted because it forced a clunky public API change for marginal benefit.
2. **Token endpoint error responses** — intentionally out of scope. RFC 6749 §5.1's MUST is "any response containing tokens, credentials, or other sensitive information." Error responses carry only `error`/`error_description`, not a token or credential, so they're not in §5.1's scope. The EMA error path keeps its headers only because that was already on `main`; I didn't want to silently expand the rule to ~40 non-sensitive error sites in this PR. Happy to do it as a separate defense-in-depth change if maintainers want it.
3. **DCR unconditional** — agreed, harmless; `no-store` on a public-client 201 is fine.
4. **Unnecessary spread (line 2911)** — fixed in 4b89e6c, now passes `NO_CACHE_HEADERS` directly. Typecheck + 484 tests green.

PR #227, opened 2026-06-05
RFC: capability-aware pluggable OAuth storage providers

mattzcarey:
## Status

Draft design RFC. This PR intentionally contains no storage implementation.

The previous KV-shaped storage abstraction was reset to `origin/main`. Its last tip is preserved locally as `archive/pr-227-storage-v1` for reference.

## Why restart

Mirroring `KVNamespace.get/put/delete/list` made the existing call sites portable, but did not make authorization-code consumption, refresh rotation, replay reservation, token issuance, or revocation atomic. It also exposed the current KV key layout as the permanent public adapter contract.

Storage guarantees are part of the OAuth security model. The provider needs to know which guarantees an adapter offers and disable or degrade features honestly.

## Proposed design

`docs/storage-providers.md` specifies:

- a provider object and request-scoped normalized connection, inspired by pi-ai's provider boundary;
- a domain-specific OAuth storage contract instead of generic KV or database CRUD;
- explicit capability descriptors and a generated compatibility report;
- fenced begin/commit/abort transitions around authorization-code and refresh-token callbacks;
- atomic grant-plus-token issuance and guarded token exchange issuance;
- typed client, grant, token, consent, replay, and maintenance stores;
- adapter-owned migrations and cleanup;
- one capability-aware adapter conformance suite;
- dependency-free core integration through injected Postgres and Redis clients.

## Planned adapters

- Workers KV: backward-compatible default with explicit best-effort/eventual guarantees.
- D1: dependency-free general-purpose SQLite adapter.
- Durable Object SQLite: strongly serialized per-grant state, with global query capabilities determined by its index strategy.
- PostgreSQL: injected client contract used through cache-disabled Hyperdrive, including PlanetScale deployments.
- Redis: injected command client; strong operations require atomic scripts or an equivalent primitive.

D1 and Durable Objects are both SQLite-backed but require separate adapters. Hyperdrive is connection pooling and optional query caching for PostgreSQL, not a database adapter. PlanetScale uses that same PostgreSQL-through-Hyperdrive path. OAuth storage must use a cache-disabled Hyperdrive binding.

## Main open decisions

1. D1 or Durable Object SQLite first for the strong adapter.
2. Durable Object global index strategy.
3. Per-record partitions versus a fixed shard set for Durable Objects.
4. Subpath exports versus separate Postgres and Redis packages.
5. Compatibility-to-strict rollout timing.
6. Transition lease renewal and callback timeout policy.
7. Refresh grace versus idempotent replay handling on strong adapters.
8. Runtime unsupported helper errors versus capability-dependent helper types.

## Verification

- `npm run check`: 519 tests pass.
- `npx prettier --check docs/storage-providers.md`: pass.
- `git diff --check`: pass.

### DIFF
--- a/docs/storage-providers.md
+++ b/docs/storage-providers.md
@@ -115,15 +115,15 @@ Both D1 and SQLite-backed Durable Objects use SQLite, but their execution models
 
 They need separate adapters even if they share SQL fragments and record schemas.
 
-### PostgreSQL and MySQL
+### PostgreSQL through Hyperdrive
 
-Hyperdrive is connection pooling and optional query caching for an existing Postgres or MySQL database. It is not the database and it is not a SQL driver.
+Hyperdrive provides connection pooling and optional query caching for an existing PostgreSQL database. It is not the database and it is not a SQL driver.
 
-The application still supplies a driver such as `pg`, Postgres.js, or `mysql2`. The OAuth storage contract should accept a small driver-neutral connection factory instead of adding one of those packages as a runtime dependency.
+The application still supplies a PostgreSQL driver such as `pg` or Postgres.js. The OAuth storage contract should accept a small driver-neutral connection factory instead of adding one of those packages as a runtime dependency.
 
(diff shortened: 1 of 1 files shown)


mattzcarey:
The RFC has been accepted as the implementation direction. Implementation will continue in a separate draft PR.

PR #231, opened 2026-06-11
perf: bound grant pagination when revoking existing grants

william-canva:
## Summary

- In `completeAuthorization`, the existing-grant revocation loop called `listUserGrants` without a `limit`, so a user with many grants could trigger thousands of KV reads in a single invocation — risking KV throttling and the Cloudflare subrequest limit.
- Cap each page fetch with an explicit batch size (default 50), mirroring the existing `DEFAULT_PURGE_BATCH_SIZE` pattern used by `purgeExpiredData`.
- Adds an optional `revokeExistingGrantsBatchSize` field to `CompleteAuthorizationOptions`. Behavior is unchanged for callers who don't set it — pagination still walks every page; only the per-page fetch size is now bounded.

## Test plan

- [x] `npm run typecheck` passes
- [x] `npm test` — full suite passes (504 tests)
- [x] New tests in the existing `WITH revokeExistingGrants (the fix)` block:
  - default batch size 50 is passed to `OAUTH_KV.list` for `grant:user-1:` listing
  - custom `revokeExistingGrantsBatchSize` is honored on the underlying `list` call
  - with 5 pre-existing grants and `revokeExistingGrantsBatchSize: 2`, pagination still revokes every old grant (1 grant remains after re-auth)

### DIFF
--- a/.changeset/bound-revoke-grants-pagination.md
+++ b/.changeset/bound-revoke-grants-pagination.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Bound the KV page size used when revoking existing grants during authorization.
--- a/README.md
+++ b/README.md
@@ -229,6 +229,15 @@ class ApiHandler extends WorkerEntrypoint {
 }
 ```
 
+By default, `completeAuthorization()` revokes existing grants for the same user and client after storing the new
+grant. This prevents stale tokens from continuing to use old `props` after a user re-authorizes. Set
+`revokeExistingGrants: false` only if your application intentionally allows multiple concurrent grants for the same
+user and client.
+
+For users with many grants, `revokeExistingGrantsBatchSize` controls the KV page size used while scanning existing
+grants for revocation. It defaults to `50`, must be a positive integer, and is capped at Cloudflare KV's maximum page
+size of `1000`.
+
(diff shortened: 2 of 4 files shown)


threepointone:
Thanks for putting this together. This is a useful improvement and worth landing/releasing: bounding the KV page size reduces the per-page fanout from `completeAuthorization()` when it scans existing grants before revocation.

I pushed a follow-up commit that tightens the new option before merge:

- validates `revokeExistingGrantsBatchSize` so `0`, negative, fractional, `NaN`, and infinite values are rejected instead of being passed through to KV
- clamps values above Cloudflare KV's `1000` key page limit
- adds regression tests for those edge cases
- documents the option in the README and adds a patch changeset

One important gap remains, and I don't think we should pretend this PR fully solves it: this bounds the KV page size, not the total amount of work per authorization. A user with thousands of grants can still make `completeAuthorization()` walk every page and read every grant in one invocation because grants are keyed by `grant:${userId}:${grantId}` and we have to inspect each record to find matching `clientId`s.

A complete fix for that likely needs a bigger design, either a secondary index by user+client or a bounded/resumable cleanup flow. I think it's still reasonable to land this PR as an incremental improvement, then handle the deeper total-work bound separately.

threepointone (approved):

PR #237, opened 2026-07-09
feat: add capability-aware OAuth storage providers

mattzcarey:
## Summary

Implements the accepted storage-provider direction from closed RFC #227 in a new branch and draft PR.

This changes storage from direct `env.OAUTH_KV` access to request-scoped, capability-aware OAuth domain operations. Workers KV remains the zero-configuration default and preserves existing physical keys, canonical JSON values, and active records.

## Architecture

New organized source tree:

- `src/storage/` — public backend-neutral contracts, capabilities, compatibility reports, lifecycle, records, errors, result unions, transitions, and pagination.
- `src/storage/kv/` — legacy-compatible Workers KV adapter.
- `src/storage/d1/` — D1 adapter and versioned migration API.
- `src/storage/durable-object/` — partitioned Durable Object SQLite adapter and `OAuthStorageObject`.
- `src/storage/postgres/` — driver-neutral PostgreSQL adapter and versioned migration API.
- `src/storage/redis/` — injected-client, Lua-CAS Redis adapter.
- `src/storage/testing/` — runner-neutral conformance cases for external adapters.

The OAuth engine owns protocol validation, cryptography, record construction, and OAuth error mapping. Adapters own physical encoding, SQL, Lua, transactions, migrations, indexes, and cleanup.

## Request lifecycle

- One storage connection is opened per request and reused by all OAuth and handler helper operations.
- Detached helpers open and close one connection per public operation.
- Request-bound helpers cannot reuse a closed connection after the handler resolves.
- Callback/handler errors remain primary when `close()` also fails.
- Both `ExportedHandler` and `WorkerEntrypoint` paths use the same request-scoped boundary.
- Production source contains no direct KV CRUD outside the KV adapter.

## Capabilities and compatibility

Each adapter publishes semantic guarantees for:

- read visibility;
- client create/replace;
- composite grant/token issuance;
- authorization-code and refresh transitions;
- replay reservation;
- access-token, grant, and client revocation;
- consent CAS;
- indexed queries;
- logical expiry and physical cleanup.

`storageGuarantees` defaults to `"compatibility"`. `"strict"` rejects construction when an enabled OAuth flow lacks strong guarantees. `getStorageCompatibility()` returns the static report without opening storage.

Unsupported helper/query operations fail before adapter I/O with a redacted `OAuthStorageError`.

### Workers KV

- Default when `storage` is omitted.
- Preserves `client:`, `grant:`, `token:`, and `enterprise-jti:` keys in the default namespace.
- Preserves canonical JSON without adding storage envelopes.
- Explicitly advertises eventual reads and best-effort transitions, replay reservation, and cascades.
- Non-default namespaces use a collision-safe prefix.

KV cannot make grant-plus-token transition commits atomic. Its exact partial-write behavior remains documented and tested rather than described as strong.

### D1

- Normalized clients, grants, tokens, consent, replay, lease, fence, and schema tables.
- D1 transactional batches with SQL assertion rows force rollback when guarded preconditions miss.
- Strong code/refresh fences, composite issuance, token guards, registered-only cascades, replay, and consent CAS.
- External/CIMD grant provenance remains independent of registered-client deletion.
- Versioned migration state rejects unsupported newer schemas.

### Durable Object SQLite

- Hashes namespace, aggregate kind, and logical key into deterministic non-secret object names.
- One object per user owns that user's grants, access tokens, consent, transition leases, and monotonic fences.
- One object per registered client provides strong client create/replace; 256 bounded replay shards per reservation namespace provide strong set-if-absent replay reservation.
- Same-user issue/replace, token mutation, grant cascade, consent CAS, and code/refresh transitions share one `transactionSync` domain.
- Aggregate metadata binds each physical SQLite object to one namespace/kind/key and rejects routing collisions or misrouted commands.
- Request-scoped connections memoize only deterministic object-name promises; the cache clears on close and never retains stubs or storage state.
- Route validation compares compact aggregate kind/key tuples before SQLite effects; typed record decoding, ordinary logical-expiry reads, and guarded record updates share deep implementations. The built DO entrypoint is 1,629 bytes smaller than the initial partitioned version.
- Logical expiry is authoritative; per-object alarms cascade expired grants to their tokens and clean local records.
- User-local grant/consent/token operations omit redundant JSON user predicates and root-era per-user indexes; client and grantId token indexes are created only for user aggregates; replay shards omit all records-table indexes.
- Fresh client/replay/user initialization performs 15.2% fewer schema calls than creating all indexes in every object; selective setup is transactional and rollback-tested.
- Registered-client validation is an honest best-effort cross-object precondition. Client listing, grants-by-client, cross-user client cascade, and global purge are unsupported before Durable Object I/O without an authoritative global index.

Trade-off: the built-in partitioned adapter uses the compatibility profile and does not satisfy `storageGuarantees: "strict"`. Use D1, PostgreSQL, or Redis for a complete strong global profile.

### PostgreSQL

- Zero bundled driver: applications inject one exclusive request session with parameterized `query()` and `release()`.
- Strong row-lock/transaction semantics for issuance, transitions, cascades, replay, consent, and maintenance.
- Nullable registered-client provenance allows external/CIMD grants while preserving registered-client FK cascades.
- Composite keyset pagination avoids duplicate/omitted grant and consent rows.
- Strict safe-integer decoding rejects oversized PostgreSQL `bigint` revisions/fences/timestamps/counts.
- Migrations serialize with a PostgreSQL advisory lock and advance versions monotonically.

OAuth storage through Hyperdrive must use a cache-disabled binding. PlanetScale PostgreSQL uses this same PostgreSQL-through-Hyperdrive path; it is not a separate adapter. MySQL is not included.

### Redis

- Zero bundled driver: applications inject authoritative-primary `get()` and atomic `eval()`.
- One hash-tagged, versioned state key per logical namespace; every mutation commits through one Lua compare-and-swap script.
- Strong composite issue, transitions, cascades, replay, consent, queries, and maintenance.
- Strict persisted-state decoding rebuilds null-prototype maps and rejects malformed schemas.
- CAS retries are bounded and malformed script results fail immediately.

Trade-off: each namespace is limited by one Redis key's size and one-key write throughput.

## Security properties

- Raw authorization codes, access tokens, refresh tokens, client secrets, and replay JTIs never cross the storage contract.
- Credential identifiers are lowercase SHA-256 digests.
- Canonical encrypted props and wrapped encryption keys retain the existing security model.
- Strong transitions persist bounded leases and monotonic fences; stale commits are rejected.
- No database transaction or row lock is held across `tokenExchangeCallback` network I/O.
- Callbacks receive a stable idempotency key for one grant/credential attempt.
- Logical expiry is authoritative on reads and guarded mutations; backend TTL/alarm cleanup is physical maintenance only.
- Storage errors serialize without keys, values, SQL, connection strings, or backend messages.

## Public package exports

- `@cloudflare/workers-oauth-provider`
- `@cloudflare/workers-oauth-provider/storage`
- `@cloudflare/workers-oauth-provider/storage/kv`
- `@cloudflare/workers-oauth-provider/storage/d1`
- `@cloudflare/workers-oauth-provider/storage/durable-object`
- `@cloudflare/workers-oauth-provider/storage/postgres`
- `@cloudflare/workers-oauth-provider/storage/redis`
- `@cloudflare/workers-oauth-provider/storage/testing`
- historical `.../dist/oauth-provider.js`

The package remains at zero runtime dependencies. A minor changeset is included.

## Verification

Local final gates:

- `npm run build` — pass.
- `npm run check:package` — pass; all 17 export targets exist, pack correctly, install into a clean fixture, and typecheck under NodeNext and Bundler resolution.
- `npm run check` — pass: 718 tests passed; service-gated integration files skipped in the default run.
- `RUN_POSTGRES_INTEGRATION=1 POSTGRES_PORT=55432 npx vitest run __tests__/storage/postgres/integration.test.ts` — 6 passed against PostgreSQL 16.
- `RUN_REDIS_INTEGRATION=1 REDIS_PORT=56379 npx vitest run __tests__/storage/redis/integration.test.ts` — 2 passed against Redis 7.4.
- `npx prettier --check .` — pass.
- `git diff --check` — pass.
- Direct production `OAUTH_KV.get/put/delete/list` grep — no matches.

CI adds required PostgreSQL 16 and Redis 7.4 service jobs so real transaction/Lua tests do not rely on local opt-in execution.

## Review focus

This is intentionally a large draft. Suggested review order:

1. `src/storage/contracts` represented by the top-level contract modules.
2. Provider lifecycle and transition integration in `src/oauth-provider.ts`.
3. Workers KV compatibility and storage-schema preservation.
4. D1 and Durable Object transaction/fence semantics.
5. PostgreSQL locks, migrations, and provenance.
6. Redis single-key CAS model and limitations.
7. Package exports, docs, and conformance tests.


chrisdoc:
**Production evidence for the KV eventual-read hazard**

I maintain hevy-mcp. The hosted worker at `mcp.hevy-mcp.dev` uses this provider, version 0.10.2, with the default `OAUTH_KV` storage.

On the Claude.ai OAuth connector, authorization succeeds. The provider issues a token. The `/mcp` request that follows immediately fails. The connector shows "returned an error when connecting." A retry after 60 to 90 seconds succeeds with the same token.

The cause is a read-after-write race on the token record. The `/token` handler writes the token to KV. The `/mcp` request then reads the same record with a single `get()`. KV propagates writes to server locations gradually. The two requests can reach different server locations. The read can miss the new record.

The provider then returns `401` with `Bearer error="invalid_token"`. The connection ends. The error is `invalid_token`, not `invalid_target`. The failure is not an audience or resource mismatch.

This PR keeps "eventual read-after-write across locations" for the KV adapter. KV is also the default storage. So an upgrade to this PR alone does not remove this failure. That decision is sound. It means that a real deployment must use the D1 or Durable Object adapters to get read-your-writes.

We plan to move our token storage to D1. We will test the D1 and the Durable Object adapters after this PR merges. We can add a conformance test. The test reproduces the token-issued-then-validated sequence.

Full details are on https://github.com/chrisdoc/hevy-mcp/issues/1039 .

PR #238, opened 2026-07-13
chore: replace internal `env: any` with a typed binding shape

### LINKED ISSUE #213: chore: replace internal `env: any` with a typed binding shape
## Problem

`oauth-provider.ts` uses `env: any` on ~25 internal method signatures and several public-ish helper signatures. Several other parameters and helper fields are `: any` as well (40 occurrences across the file). This pushes type unsafety down the call graph — every site that touches `env.OAUTH_KV.get(...)`, `env.OAUTH_KV.put(...)`, etc. silently bypasses TypeScript's binding checks, and any future binding (D1, R2, Queues, secrets) is invisible to readers.

Examples:

```ts
// src/oauth-provider.ts
async unwrapToken<T = any>(token: string, env: any): Promise<TokenSummary<T> | null>
private async handleAuthorizationCodeGrant(body: any, clientInfo: ClientInfo, env: any): Promise<Response>
private async handleRefreshTokenGrant(body: any, clientInfo: ClientInfo, env: any): Promise<Response>
private async handleTokenExchangeGrant(body: any, clientInfo: ClientInfo, env: any): Promise<Response>
private async revokeToken(body: any, env: any): Promise<Response>
private async validateAccessToken(tokenId: string, userId: string, grantId: string, env: any): Promise<boolean>
// ... and ~20 more
```

The `OAuthProvider<Env>` class already takes an `Env` generic. Internal methods should thread it through instead of erasing it to `any`.

## What "correct" looks like

The library's own contract is that `OAUTH_KV: KVNamespace` is required. Everything else is the deployer's app-specific bindings the library doesn't touch. The minimal correct shape on every internal `env` parameter is something like:

```ts
type ProviderEnv = { OAUTH_KV: KVNamespace };

private async revokeToken(body: BodyShape, env: ProviderEnv): Promise<Response>
```

For public callbacks that surface the deployer's env (`tokenExchangeCallback`, `EmaClaimsMapper`, `EmaTrustedIssuerResolver`, the `apiHandler`'s `env` param, etc.), the library should already be threading the `Env` generic through — fix the spots where it isn't.

## Scope

Single-PR refactor. No behavior change. Touches `src/oauth-provider.ts` primarily and a couple of EMA call sites (`runEmaPipeline`, `issueEmaAccessToken`) that currently take `env: any` from the parent handler.

The `body: any` problem is adjacent (parsed token-request body) but separate — `body` legitimately comes from an untrusted form/JSON parse and has no static schema. The fix there would be a body-validation pass before any handler sees it. Out of scope for this issue.

## Why this matters

- Future binding additions (D1, R2, Queue, Hyperdrive, Vectorize) silently work without type checks
- Tests pass `mockEnv` shapes that don't have to match production bindings
- A typo in a binding name (`env.OAUTH_KV` vs `env.OUATH_KV`) only fails at runtime
- The `EmaJtiStore.markUsed` `env` field was recently tightened to `{ OAUTH_KV: KVNamespace }` ([example](https://github.com/cloudflare/workers-oauth-provider/blob/main/src/ema/types.ts)) — the rest of the file should follow suit

thinkter:
Fixes #213.

## What

Removes all 24 `env: any` occurrences from `src/oauth-provider.ts` and replaces them with typed shapes, per the issue:

- Adds an internal `ProviderEnv` interface declaring the one binding the library requires (`OAUTH_KV: KVNamespace`), mirroring the shape already used by `EmaJtiStore.markUsed` in `src/ema/types.ts`.
- Every internal method of `OAuthProviderImpl<Env>` (token/grant handlers, revocation, registration, API dispatch, `saveGrantWithTTL`, `getClient`, `unwrapToken`, `exchangeToken`, and the EMA `runEmaPipeline` / `issueEmaAccessToken` arg objects) now takes `env: Env & ProviderEnv` instead of `env: any` — so `env.OAUTH_KV` access is type-checked *and* the deployer's full `Env` still flows untouched into `apiHandler.fetch`, the default handler, `tokenExchangeCallback`, and the EMA callbacks.
- `ResolveExternalTokenInput` is now generic (`ResolveExternalTokenInput<Env = Cloudflare.Env>`) and `OAuthProviderOptions.resolveExternalToken` threads `Env` through, so the callback sees the deployer's typed env. The default type parameter keeps existing consumers compiling unchanged.
- `OAuthHelpersImpl` is generic over `Env` instead of holding `env: any` / `OAuthProviderImpl<any>`.
- `CreateAccessTokenOptions.env` takes the minimal `ProviderEnv` shape, since `createAccessToken` only writes to KV.

## What deliberately didn't change

- **Public API surface:** `OAuthProvider.fetch` / `purgeExpiredData` and `getOAuthApi` still accept a plain `Env` (no new `extends` constraint, which would be a breaking change for TS consumers whose `Env` types don't declare `OAUTH_KV`). Each performs a single `env as Env & ProviderEnv` cast at the boundary — the same runtime contract the library has always had.
- **`body: any`:** out of scope per the issue; it needs a body-validation pass, not a type annotation.
- **No runtime behavior change.** The diff is type annotations plus the boundary casts only.

## Testing

- `npm run check` (tsc + vitest): 519/519 tests pass
- `npm run build` (tsdown): clean
- `npx prettier -c .`: clean
- Changeset included (patch)


mattzcarey:
Thanks for contributing this :) I recreated it as #271 under the Cloudflare repository so the branch gets the normal CI path, and preserved your implementation commit and authorship there. Superseded by #271.

thinkter:
thanks alot! anything else that I can work on?

PR #248, opened 2026-07-27
fix(cimd): report metadata fetch failures instead of treating them as unknown clients

ksinder:
## What changed

CIMD metadata resolution failures are now distinguishable from a client that does not exist.

Previously, `getClient()` caught every failed Client ID Metadata Document fetch or validation and returned `null`. Timeouts, WAF blocks, HTTP failures, oversized responses, and invalid metadata therefore looked identical to an unregistered client.

The provider now throws an exported `CimdFetchError` with:

- `reason: "metadata_resolution_failed"` — stable for telemetry and control flow
- `metadataUrl` — the CIMD URL
- `detail` — the underlying diagnostic message

### Token endpoint

The wire response remains the same generic `401 invalid_client` / `Client not found` response. The provider:

- preserves the RFC 6749 Basic challenge when authentication was attempted through HTTP Basic
- reports `internal.category: "client-id-metadata-document"`
- reports the stable `internal.reason: "metadata_resolution_failed"`
- puts the metadata URL and dynamic message in `internal.detail`
- includes the originating request for request-scoped correlation

### OAuth helpers

`OAuthHelpers.lookupClient()`, `parseAuthRequest()`, `completeAuthorization()`, and `exchangeToken()` let `CimdFetchError` propagate. `lookupClient()` now returns `null` only when the client does not exist. The public interface and README document the exception contract and recovery pattern.

This is a minor change because callers that previously treated CIMD resolution failures as `null` must catch `CimdFetchError` to retain that behavior.

## Standards basis

The Client ID Metadata Document draft says authorization servers should abort authorization when metadata retrieval fails and must not cache error responses or invalid documents. It does not prescribe a specific token-endpoint error body, so the provider keeps the existing generic wire response while exposing diagnostics only to trusted application code.

## Tests

Coverage includes:

- network, HTTP, timeout, size-limit, and metadata-validation failures
- stable error and telemetry fields
- generic token-endpoint wire responses
- Basic challenge preservation
- no Basic challenge for form-post authentication
- current CIMD authentication-method and metadata validation behavior

Validated with Node 24:

- `npm run check` — 558 tests
- `npm run build`
- `npx prettier --check .`
- `npx changeset status`
- `git diff --check`

### DIFF
--- a/.changeset/cimd-fetch-error-propagation.md
+++ b/.changeset/cimd-fetch-error-propagation.md
@@ -7,9 +7,10 @@ A failed Client ID Metadata Document fetch previously became a `null` client
 lookup, so a network problem (timeout, WAF block, upstream outage) was
 indistinguishable from an unregistered client — at the token endpoint, in the
 `onError` hook, and for `OAuthHelpers` callers. The fetch failure now throws a
-new exported `CimdFetchError` carrying the metadata URL and underlying reason.
-The token endpoint still returns the same generic `invalid_client` / "Client
-not found" response, but reports the failure through the `onError` hook's
+new exported `CimdFetchError` carrying the metadata URL, stable
+`metadata_resolution_failed` reason, and underlying diagnostic detail. The
+token endpoint still returns the same generic `invalid_client` / "Client not
+found" response, but reports the failure through the `onError` hook's
 `internal` field (category `client-id-metadata-document`) together with a new
 optional `request` field. Breaking for callers of `OAuthHelpers.lookupClient`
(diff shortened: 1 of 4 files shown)


mattzcarey (dismissed):
fixing a few bits. Thanks for this :)

mattzcarey (approved):
Re-reviewed on current main after the review-fix commit. Basic challenge handling, stable internal telemetry, public exception contracts, README guidance, and current CIMD validation tests are all aligned. Local Node 24 validation passes with 569 tests, build, Prettier, changeset status, and diff check.

PR #249, opened 2026-07-28
Preserve subject token scope during token exchange

mattzcarey:


mattzcarey:
@agent-think review this pls

PR #256, opened 2026-07-29
feat: derive resource policy from protected resource metadata

mattzcarey:
## Summary

Use `resourceMetadata.resource` as the resource-policy signal instead of adding a separate `requireResourceIndicator` option.

When `resourceMetadata.resource` is configured:

- authorization requests must contain that one exact resource
- authorization-code, refresh-token, and token-exchange requests must repeat that exact resource
- `completeAuthorization()` re-applies the policy before creating a grant
- internal and externally resolved tokens must have that exact audience
- `resourceMatchOriginOnly` cannot weaken the configured policy

When no resource is configured:

- valid RFC 8707 resource indicators are accepted
- token and refresh requests may omit `resource` and inherit it from the authorization grant
- an authorization request that omits `resource` uses its origin as the RFC 8707 default, so newly issued tokens remain origin-bound
- token requests cannot add a resource that was absent from the authorization grant

Audience validation now also preserves a query component when the resource identifier includes one.

## Standards basis

The [MCP 2026-07-28 authorization specification](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization#resource-parameter-implementation) requires MCP clients to include `resource` in authorization and token requests and requires MCP servers to validate that tokens were issued for them.

[RFC 8707 Section 2.1](https://www.rfc-editor.org/rfc/rfc8707.html#section-2.1) also allows an authorization server to process an omitted authorization resource with a predefined default. The request origin provides an interoperable default for co-located authorization and resource servers. Split deployments and deployments needing path-level isolation can opt into exact behavior by configuring the canonical resource.

## Consumer review

Checked current consumers before choosing the configuration signal:

- `cloudflare/mcp` supplies `resourceMetadata.resource_name`, but not `resource`; its existing omission-compatible flow remains on the default path.
- `cloudflare/mcp-server-cloudflare` does not configure a resource and remains on the default path.
- `getsentry/sentry-mcp` owns path-specific protected-resource metadata, sends a resource at authorization, and omits it during code exchange and refresh. The provider continues to inherit that resource from the grant.

Using the whole `resourceMetadata` object as the signal would have broken `cloudflare/mcp`, so only the `resource` field enables exact policy.

## Migration note

This changes the meaning of an existing field. Deployments that already set `resourceMetadata.resource` must ensure clients send that exact value in both authorization and token requests. External token resolvers must return the same audience. This is a minor release.

## Tests

Added coverage for:

- origin-bound defaults when clients omit `resource`
- authorization resource inheritance at code exchange
- prevention of token-request resource upscoping
- exact configured-resource enforcement at authorization, code exchange, refresh, token exchange, and `completeAuthorization()`
- rejection of legacy broad internal audiences under a path-specific policy
- internal and external audience validation
- configured query-component preservation
- metadata-only configuration remaining permissive

Validated with Node 22:

- `npm run check` (557 tests)
- `npm run build`
- `npx prettier --check .`

### DIFF
--- a/.changeset/tough-birds-resource.md
+++ b/.changeset/tough-birds-resource.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': minor
+---
+
+Use `resourceMetadata.resource` as the strict RFC 8707 resource policy. Configured deployments now require that exact resource throughout authorization, token issuance, and access-token validation. Unconfigured deployments remain interoperable by inheriting requested resources and defaulting omitted authorization resources to the request origin.
--- a/README.md
+++ b/README.md
@@ -527,13 +527,20 @@ Call it repeatedly via a cron trigger — deleted records disappear from KV, so 
 
 ## Protected Resource Metadata (RFC 9728)
 
-The library automatically serves a `/.well-known/oauth-protected-resource` endpoint. By default, it uses the request origin as the resource identifier and the token endpoint's origin as the authorization server. You can customize this with the `resourceMetadata` option:
+The library automatically serves a `/.well-known/oauth-protected-resource` endpoint. By default, it derives the resource identifier from the metadata request and the authorization server from the token endpoint.
+
+Resource policy follows the configuration:
+
+- When `resourceMetadata.resource` is configured, authorization requests and token requests must contain that one exact resource. Issued and externally resolved access tokens must have that exact audience.
+- Without an explicitly configured resource, valid RFC 8707 resource indicators are accepted. Token and refresh requests may omit `resource` and inherit it from the authorization grant. If the authorization request also omits it, the provider uses the request origin as the RFC 8707 default and issues an origin-bound token.
+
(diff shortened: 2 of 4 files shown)


mattzcarey:
Validated this against the real `cloudflare/mcp` consumer in https://github.com/cloudflare/mcp/pull/183.

The consumer installs the pkg.pr.new build from this PR, configures `resourceMetadata.resource` as the canonical `/mcp` endpoint, and passes its full local suite (292 tests). Its CI, CodeQL, and Semgrep checks are also green.

Deployed the consumer branch to staging:

- Worker: `cloudflare-api-mcp-staging`
- URL: `https://staging.mcp.cloudflare.com`
- Version: `8316f50e-d0bb-4199-a635-be7d31b1d498`

Deployed smoke checks passed:

- protected resource metadata returns `https://staging.mcp.cloudflare.com/mcp`
- authorization server metadata advertises RFC 9207 and the expected issuer
- the protected endpoint returns the bare path-specific bearer challenge
- the exact configured resource reaches consent
- broader and omitted resources are rejected before consent
- temporary DCR state was removed from staging KV after the test

From the implementation and consumer-validation side, this is ready to merge once the required review is supplied.

PR #98, opened 2025-10-27
Enhance redirect URI scheme validation for security

threepointone:
Added a robust helper to validate redirect URI schemes, preventing dangerous pseudo-schemes (e.g., javascript:, data:, vbscript:) with normalization and case-insensitive checks. Expanded test coverage to include bypass attempts using mixed case, whitespace, control characters, and edge cases to ensure comprehensive protection against XSS and related attacks.

Fixes #97 

---

I generated this with cursor agent, with the prompt: "can you solve this issue? https://github.com/cloudflare/workers-oauth-provider/issues/97  add tests as well". I then went over every line. The changeset/commit message was generated by copilot in github desktop.

PR #257, opened 2026-07-29
feat: add MCP redirect URI security policy

mattzcarey:
## Summary

Add opt-in `enforceMcpRedirectUriSecurity` for the MCP 2026 redirect transport policy.

When enabled, redirect URIs must be either:

- HTTPS, or
- HTTP loopback (`localhost`, `127.0.0.0/8`, or `::1`)

The policy is applied consistently to Dynamic Client Registration, programmatic clients, authorization requests, and Client ID Metadata Documents.

The option defaults to `false` so non-MCP native OAuth applications can continue using private-use URI schemes such as `myapp://`.

## Tests

- reject remote HTTP and private-use schemes in MCP mode
- accept HTTPS and IPv4/IPv6/localhost loopback HTTP
- preserve existing exact matching and flexible loopback-port behavior
- preserve non-MCP custom scheme compatibility

Validated with Node 24:

- `npm run check` (532 tests)
- `npm run build`
- `npx prettier --check .`

### DIFF
--- a/.changeset/safe-cats-redirect.md
+++ b/.changeset/safe-cats-redirect.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': minor
+---
+
+Add `enforceMcpRedirectUriSecurity` to restrict registered, programmatic, and CIMD redirect URIs to HTTPS or HTTP loopback addresses. This provides the redirect transport policy required by the MCP authorization security specification while preserving private-use URI schemes for non-MCP deployments by default.
--- a/README.md
+++ b/README.md
@@ -251,6 +251,20 @@ Note that `deleteClient()` cascades: it revokes all grants (and their associated
 
 See the `OAuthHelpers` interface definition for full API details.
 
+### MCP redirect URI security
+
+MCP deployments can restrict redirect URIs to HTTPS or HTTP loopback addresses across DCR, programmatic clients,
+and Client ID Metadata Documents:
+
+```ts
+new OAuthProvider({
+  // ... other options ...
+  enforceMcpRedirectUriSecurity: true,
+});
+```
+
+This option defaults to `false` so non-MCP native OAuth applications can continue using private-use URI schemes.
+
 ## Token Exchange Callback
 
 This library allows you to update the `props` value during token exchanges by configuring a callback function. This is useful for scenarios where the application needs to perform additional processing when tokens are issued or refreshed.
--- a/__tests__/oauth-provider.test.ts
+++ b/__tests__/oauth-provider.test.ts
@@ -1862,6 +1862,65 @@ describe('OAuthProvider', () => {
       });
     });
 
+    describe('MCP redirect URI security', () => {
+      const strictProvider = new OAuthProvider({
+        apiRoute: ['/api/'],
+        apiHandler: TestApiHandler,
+        defaultHandler: testDefaultHandler,
+        authorizeEndpoint: '/authorize',
+        tokenEndpoint: '/oauth/token',
+        clientRegistrationEndpoint: '/oauth/register',
+        enforceMcpRedirectUriSecurity: true,
+      });
+
+      it.each(['http://remote.example.com/callback', 'myapp://callback'])('should reject %s', async (redirectUri) => {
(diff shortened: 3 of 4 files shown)


mattzcarey:
Making this a draft and shelving it for now.

I checked this against the current MCP specification, using the exact 2026-07-28 pages:

- [Authorization](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization)
- [Authorization Server Discovery](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/authorization-server-discovery)
- [Client Registration](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/client-registration)
- [Authorization Security Considerations](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/security-considerations)

The relevant requirements are clear:

> All redirect URIs **MUST** be either `localhost` or use HTTPS.

The same security section says authorization servers **MUST** validate exact redirect URIs against pre-registered values. The client registration page shows both `http://127.0.0.1:3000/callback` and `http://localhost:3000/callback`, consistent with the OAuth 2.1 loopback exception. The authorization server discovery page does not add another redirect URI policy or an exception for private-use schemes.

So for an MCP authorization server, allowing `cursor://` or another private-use scheme is not compliant with the current MCP profile. The reason not to switch this on globally is compatibility rather than ambiguity in the MCP spec.

There is clear history for that compatibility concern in this repository:

- #80 changed redirect URIs to HTTP(S)-only.
- #87 reverted that five days later because it broke callbacks such as `cursor://`.
- #98 then made private-use schemes part of the tested behavior, including `myapp://` and `com.example.app://...`.
- #145 and #169 added loopback port flexibility and `localhost` support for native MCP clients.

Private-use redirect schemes are also supported for generic native OAuth clients by RFC 8252 and OAuth 2.1 when they use the required reverse-domain form. That means the current package behavior has real users and standards history behind it, even though it does not meet the narrower MCP 2026-07-28 profile.

There are also a few implementation gaps to fix before reviving this change:

1. `completeAuthorization()` does not run the new policy validator. It only checks that the requested URI matches the stored client metadata. A legacy client containing `cursor://...`, or a caller passing an `AuthRequest` directly without `parseAuthRequest()`, can still receive an authorization code redirect while strict mode is enabled. I reproduced this locally.

2. `updateClient()` does not validate updated `redirectUris`. A strict provider can create a client with HTTPS, update it to `cursor://...`, and persist that value successfully. I reproduced this as well.

3. Loopback port flexibility currently applies to HTTPS loopback URIs too. For example, registering `https://localhost:8443/callback` also accepts `https://localhost:9443/callback`. The port exception should only apply to HTTP loopback redirects. HTTPS redirects should continue to require an exact match.

4. Strict mode still accepts redirect URIs containing fragments, such as `https://client.example/callback#fragment`. OAuth 2.1 prohibits fragments in registered redirect URIs.

5. The new tests only exercise DCR. A complete version should cover DCR, CIMD, `createClient()`, `updateClient()`, `parseAuthRequest()`, direct `completeAuthorization()`, HTTP loopback port matching, HTTPS exact matching, and fragment rejection.

The current branch passes all 532 tests, typechecking, build, and Prettier, but it is now conflicting with `main` and will need a rebase if we return to it.

I do not think we should keep adding independent MCP compliance booleans. If this library is going to treat MCP as its protocol profile, the cleaner eventual direction is to enforce the MCP 2026-07-28 behavior by default, probably in a breaking release with a migration plan. If generic OAuth remains a supported mode, one versioned profile option such as `profile: 'mcp-2026-07-28' | 'oauth-2.1'` would be clearer than separate switches for redirect security, resource indicators, PKCE, and each future MCP requirement.

For now, the spec supports the stricter behavior, but the compatibility impact needs more thought before we turn it on for existing deployments.

mattzcarey:
Parked until after 1.0.

mattzcarey:
Superseded by #340, which shipped the MCP redirect URI policy on by default (https or loopback http, no userinfo or fragment, private-use schemes behind `allowPrivateUseRedirectUris`), enforced at DCR, CIMD, `createClient`, `updateClient` and `parseAuthRequest`.

PR #271, opened 2026-07-29
chore: replace internal `env: any` with typed bindings

### LINKED ISSUE #213: chore: replace internal `env: any` with a typed binding shape

mattzcarey:
Supersedes #238 and preserves @thinkter's implementation commit and authorship. Thanks to @thinkter for the contribution.

Closes #213.

## Summary

Replace internal `env: any` usage with the provider's required binding shape while preserving the deployer's complete `Env` type.

- add an internal `ProviderEnv` shape containing `OAUTH_KV: KVNamespace`
- thread `Env & ProviderEnv` through token, grant, registration, revocation, API dispatch, and EMA internals
- make `OAuthHelpersImpl` generic over the deployer's `Env`
- make `ResolveExternalTokenInput<Env>` generic so `resolveExternalToken({ env })` exposes application bindings without casts
- retain the unconstrained public `OAuthProvider.fetch`, `purgeExpiredData`, and `getOAuthApi` signatures, with one boundary cast matching the existing runtime contract
- leave untrusted token-request `body: any` parsing for a separate validation change

There is no runtime behavior change.

## Type regression coverage

Added a compile-time and runtime test proving that a custom provider environment reaches `resolveExternalToken` with its exact type. The test fails against `main` and passes with this branch.

## Downstream check

Packed this branch and tested it against the current `cloudflare/mcp` consumer PR:

- `cloudflare/mcp` passes its complete 292-test suite
- its stale OAuth `@ts-ignore` can be deleted and typechecking still passes
- the remaining `cloudflareEnv as AuthEnv` assertion represents runtime injection of `OAUTH_PROVIDER` into a globally imported binding; this internal-env refactor cannot remove that assertion

## Validation

- `npm run check` (573 tests)
- `npm run build`
- `npx prettier --check .`
- `cloudflare/mcp`: `npm run check` (292 tests)

### DIFF
--- a/.changeset/typed-internal-env.md
+++ b/.changeset/typed-internal-env.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Replace internal `env: any` parameters with typed shapes. Internal provider methods now take `Env & ProviderEnv` (where `ProviderEnv` declares the required `OAUTH_KV: KVNamespace` binding), `ResolveExternalTokenInput` threads the `Env` generic through to the `resolveExternalToken` callback, and `OAuthHelpersImpl` is generic over `Env`. No runtime behavior change.
--- a/__tests__/oauth-provider.test.ts
+++ b/__tests__/oauth-provider.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach, vi, afterEach } from 'vitest';
+import { describe, it, expect, expectTypeOf, beforeEach, vi, afterEach } from 'vitest';
 import {
   CimdFetchError,
   OAuthError,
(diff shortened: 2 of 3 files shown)


thinkter:
tysm! 
is there anything else that I can work on
would love to help!

PR #298, opened 2026-08-10
feat(metadata): add opt-in issParameterCompat to omit the RFC 9207 advertisement

mattzcarey:
## Motivation

Staging results for 0.10.2: Claude.ai, Claude Code, Cursor, and ChatGPT CIMD all pass. Codex CLI 0.143–0.146 fails because it drops the RFC 9207 `iss` callback parameter before validating it (openai/codex#31573). 0.10.2 exposes this by correctly advertising `authorization_response_iss_parameter_supported: true` — Codex's OAuth stack sees the advertisement, expects to validate `iss`, and fails on its own dropped parameter.

## Change

New opt-in provider option `issParameterCompat` (default `false`):

- When `true`, the RFC 8414 authorization server metadata **omits** `authorization_response_iss_parameter_supported` entirely (absent, not `false`).
- The `iss` parameter itself is **still sent on every authorization response** (code-flow query param and implicit-flow fragment), unconditionally — this flag never touches it. Clients that implement RFC 9207 validation keep full authorization-server mix-up protection.
- Default behavior is byte-identical: same key, same position in the serialized metadata.

The JSDoc marks it as a temporary compatibility flag to be removed once affected clients update.

## Note on Codex versions

The staging report said this was fixed in Codex 0.147, but upstream openai/codex#31573 was still open with no fix release at the time of writing, so the docs say "0.143 and later, at the time of writing" and cite the issue. Happy to tighten to a version range if you've verified 0.147.

## Testing

- New test: with the flag set, metadata has no `authorization_response_iss_parameter_supported` property (mutation-checked: reverting the metadata line fails this test).
- New test: with the flag set, a completed authorization redirect still carries `iss=https://example.com` (and a real `code`).
- Existing default-behavior tests (advertised `true`) untouched and passing.
- `npm run check`: tsc clean, 850/850 tests, prettier clean.

Includes a patch changeset.

### DIFF
--- a/.changeset/iss-parameter-compat.md
+++ b/.changeset/iss-parameter-compat.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': patch
+---
+
+Add an opt-in `issParameterCompat` option that omits the RFC 9207 `authorization_response_iss_parameter_supported` advertisement from authorization server metadata for clients that mishandle it (e.g. Codex CLI 0.143 and later, which drops the `iss` callback parameter before validating it — see openai/codex#31573). The `iss` authorization-response parameter is still always sent, so conforming clients keep authorization-server mix-up protection. Defaults to false.
--- a/README.md
+++ b/README.md
@@ -428,6 +428,7 @@ Deleting a client through `OAuthHelpers.deleteClient()` also revokes its grants 
 | `clientIdMetadataDocumentEnabled`  | Enable CIMD lookup and advertisement                     | `false`                                     |
 | `allowPlainPKCE`                   | Permit the legacy plain PKCE method                      | `false`                                     |
 | `allowImplicitFlow`                | Enable implicit token responses                          | `false`                                     |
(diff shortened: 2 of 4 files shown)


mattzcarey:
Closing as unnecessary: the iss regression is fixed upstream. Verified three ways:

- openai/codex#31573 confirms the fix shipped in codex-cli 0.147.0-alpha.1 (independently verified by users on 2026-07-30).
- Source check: at tags rust-v0.147.0-alpha.1 and rust-v0.147.0-alpha.1.2, perform_oauth_login.rs parses iss from the callback query and calls handle_callback_with_issuer; 0.146.1 lacks this. Stable rust-v0.147.0 released 2026-08-07.
- Staging: the ChatGPT-bundled codex-cli 0.147.0-alpha.1.2 completed a fresh OAuth flow against strict staging with authorization_response_iss_parameter_supported advertised.

Staging is restored to stock 0.10.2. Leaving the branch in place in case a compat escape hatch is ever needed for another client.

PR #302, opened 2026-08-13
feat: support private_key_jwt for CIMD clients

### LINKED ISSUE #264: feat: support private_key_jwt for CIMD clients
## Summary

Support `private_key_jwt` client authentication for clients identified by a Client ID Metadata Document.

## Current behavior

CIMD support accepts only `token_endpoint_auth_method: "none"`. Documents that advertise both `none` and `private_key_jwt` can work because the provider selects `none`, but a client that requires `private_key_jwt` cannot complete authorization code exchange.

This is not a core MCP compliance blocker because the MCP specification makes `private_key_jwt` optional. It is useful for clients that want stronger authentication than an unauthenticated public client.

## Acceptance criteria

- Parse and validate the client's `jwks` or `jwks_uri` metadata according to the CIMD draft.
- Authenticate token endpoint requests using `private_key_jwt`.
- Validate assertion issuer, subject, audience, signature, expiration, and unique identifier/replay constraints.
- Apply SSRF protections to remotely fetched JWKS documents.
- Advertise `private_key_jwt` only when the complete token-endpoint path is supported.
- Add positive and negative tests for key rotation, invalid audience, expiry, replay, unknown keys, malformed assertions, and SSRF-sensitive URLs.
- Document how this interacts with the `global_fetch_strictly_public` compatibility flag.

## References

- https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization/client-registration#client-id-metadata-documents
- https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00
- https://datatracker.ietf.org/doc/html/rfc7523

ting-hong-shieh:
## Summary

- Adds `private_key_jwt` token endpoint authentication for CIMD clients that do not offer `none`.
- Accepts one inline `jwks` or HTTPS `jwks_uri` and verifies `RS256` and `ES256` client assertions.
- Validates `iss`, `sub`, `aud`, `exp`, optional `iat` and `nbf`, and single-use `jti` values. Replay markers are hashed and retained in `OAUTH_KV` until the assertion expires.
- Preserves the existing public-client path when a CIMD document offers both `none` and `private_key_jwt`, including the current ChatGPT metadata shape.
- Advertises `private_key_jwt` and its signing algorithms only when CIMD and `global_fetch_strictly_public` are both enabled.

Closes #264

## Security boundaries

- Rejects mixed client authentication methods, unsupported JOSE extensions, private or symmetric JWK material, and ambiguous key selection.
- Limits client assertions to 16 KB and remote JWKS responses to 64 KB with a 10 second timeout.
- Requires HTTPS for remote JWKS and relies on the existing `global_fetch_strictly_public` requirement to block private-network fetches.
- Uses the same best-effort KV replay model as the existing EMA assertion path; simultaneous requests in different colos remain subject to KV's eventual consistency.

## Validation

- `npm run check`: 9 test files, 867 tests passed
- `npm run build`
- Prettier on all changed source, test, documentation, and changeset files
- `git diff --check`

### DIFF
--- a/.changeset/strong-carrots-sign.md
+++ b/.changeset/strong-carrots-sign.md
@@ -1,0 +1,5 @@
+---
+'@cloudflare/workers-oauth-provider': minor
+---
+
+Support `private_key_jwt` authentication for Client ID Metadata Document clients, including inline and remote JWKS validation, signed client assertions, and replay protection.
--- a/README.md
+++ b/README.md
@@ -306,11 +306,13 @@ CIMD validation follows [draft-ietf-oauth-client-id-metadata-document-00](https:
 - Exact authorization-request redirect URI validation, with RFC 8252 loopback port handling.
 - A 5 KB response size limit and a 10 second timeout covering both headers and body.
 - Valid UTF-8 JSON object syntax and safe URI schemes for client metadata fields.
-- No embedded client secrets or private JWK material.
+- No embedded client secrets or private or symmetric JWK material.
 
 Validated documents are cached according to their `Cache-Control` headers, capped at 7 days. Error responses and invalid documents are never cached, and a cached document that stops validating is evicted and re-resolved from origin within the same request.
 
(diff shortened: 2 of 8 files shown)


mattzcarey:
Parking this after 1.0 rather than closing it: `private_key_jwt` for CIMD clients is still wanted, but we haven't scheduled it. I've converted it to a draft so it doesn't look ready to merge. When we pick it up, it will need rebasing onto 1.x, where client authentication failures now carry `onError.internal` reasons (#327).

Issue #214, opened 2026-05-29
Refresh-token rotation state on KV is racy under concurrent refresh; consider DO-backed serialization or idempotent rotation


ksinder:
> _Filed by @ksinder (Notion) via Claude Code, Anthropic's CLI assistant. Happy to prototype whichever direction maintainers prefer._

## Summary

Refresh-token rotation state lives on Workers KV, which is eventually consistent and has no atomic compare-and-swap. That's fine for the library's own one-step grace window, but anyone layering **reuse detection with grant revocation** on top (per RFC 9700 §4.14.2) hits a consistency hazard under concurrent refresh.

## Background

The library rotates refresh tokens with a one-step grace window: the `grant:${userId}:${grantId}` record holds `refreshTokenId` + `previousRefreshTokenId`, updated via a single read-modify-write `OAUTH_KV.put()` with no locking or CAS. Using either token mints a new one. There is no built-in reuse/replay detection — reuse of the already-rotated previous token is silently honored.

## Problem

To get RFC 9700 §4.14.2 replay detection you revoke the grant when the already-rotated previous token is replayed. But two legitimate situations replay the previous token without any attack:

- a client firing two concurrent refreshes, and
- a client retrying a refresh after a network timeout when the first attempt actually succeeded server-side.

Telling benign concurrency from theft requires reading rotation state that KV may not have propagated yet (propagation can take meaningful time, and there's no cross-key ordering), and the grant's read-modify-write can be clobbered by a concurrent writer. The result is either **over-revocation** (disconnecting healthy clients on benign retries) or missed detection.

A time-bounded grace window (the Okta/Auth0 "rotation leeway" approach) mitigates the over-revocation, but if the freshness timestamp is stored in a **separate** KV key from the grant, a concurrent replay can observe the rotated grant before it observes the freshness marker — reintroducing the race. The window concept is sound; the separate-key, eventually-consistent storage is the footgun.

Related prior discussion: #43 (rotation vs. availability tradeoff) and #149 (strict rotation for public clients, closed as a breaking change).

## Questions / possible directions

1. **Intent:** Is reuse/replay detection intended to be in scope for this library, or is grace-window-only the deliberate design? (README implies the latter; confirming would help downstream implementers who add detection themselves.)
[...]

PR #312, opened 2026-09-11
feat(storage): pluggable storage providers with Workers KV and Durable Object adapters

### LINKED ISSUE #232: Bound total KV work when revoking existing grants during authorization
## Problem

PR #231 improves `completeAuthorization()` by adding `revokeExistingGrantsBatchSize`, which bounds the number of grant keys fetched per KV `list()` page while revoking existing grants for the same user+client.

That is useful, but it does **not** fully solve the operational problem that motivated the PR: a user with many grants can still cause one authorization request to scan every grant for that user and perform one KV read per grant summary.

Today, grants are keyed as:

```text
grant:${userId}:${grantId}
```

To revoke old grants for the same user+client, `completeAuthorization()` calls `listUserGrants(userId)`, pages through `grant:${userId}:`, reads each grant record, then filters by `clientId` in application code.

That means total work is still O(number of grants for user), not O(number of grants for user+client), and it can still exceed Worker/KV subrequest budgets for high-cardinality users. PR #231 only reduces per-page fanout; it does not bound total work per invocation.

## Why this matters

`completeAuthorization()` runs in the interactive authorization path. If a user has thousands of grants, one re-authorization can still:

- perform thousands of KV reads in a single Worker invocation
- risk Cloudflare subrequest limits
- increase latency on the authorization redirect path
- potentially fail before old grants are revoked, leaving stale tokens valid

This is especially relevant because `revokeExistingGrants` defaults to `true` to prevent stale `props` from causing re-auth loops.

## Current behavior after PR #231

Improved:

- each KV list page defaults to 50 keys
- custom page sizes are validated as positive integers
- page sizes above KV's 1000-key maximum are clamped

Still not solved:

- the loop still walks all pages for `grant:${userId}:`
- each grant key still requires a KV `get()` to inspect `clientId`
- revocation candidates are collected before the new grant is written, then all matching old grants are revoked afterward
- a large number of old matching grants can still trigger many revocations in one invocation

### Option 1: Add a secondary user+client grant index

Maintain an additional key per grant that can be listed by user+client, for example:

```text
grant-client-index:${userId}:${clientId}:${grantId}
```

Then `completeAuthorization()` can list only grants for the exact user+client pair and avoid scanning grants for other clients.

Things to design:

- write index entries whenever grants are created
- delete index entries when grants are revoked or purged
- handle index entries whose grant record has expired or is missing
- decide whether index entries need TTLs matching grant TTLs
- preserve behavior for existing grants created before the index exists

Migration/backward compatibility options:

- best-effort fallback to the old user-wide scan only when no index entries are found
- dual path for one release: scan old format and create/delete index entries opportunistically
- accept that old pre-index grants may need purge/cleanup before the bound is fully effective

Because this changes KV storage shape, release planning should consider the repo guidance around schema changes and token/grant compatibility.

### Option 2: Make cleanup bounded and resumable

Keep the current key shape, but cap the amount of scan/revocation work done during one `completeAuthorization()` call and continue cleanup later.

Possible mechanisms:

- store a pending cleanup cursor/job in KV
- expose a helper method or scheduled task to continue cleanup
- use `ctx.waitUntil()` where available for post-response cleanup, while still respecting invocation/subrequest limits
- make the authorization path create the new grant promptly, then perform bounded best-effort old-grant revocation

Things to design:

- stale tokens may remain valid until cleanup catches up
- failure/retry semantics for cleanup jobs
- idempotency across repeated authorizations
- observability/result reporting for incomplete cleanup

### Option 3: Change grant storage to make user+client uniqueness direct

If the intended invariant is at most one active grant per user+client, consider storing current grants under a deterministic user+client key instead of random grant IDs, or keeping a current-grant pointer.

This is likely the largest behavior/storage change and needs careful analysis because existing tokens encode grant IDs and refresh token behavior depends on grant records.

## Acceptance criteria

A complete fix should make `completeAuthorization()` revocation work bounded in total, not only per KV page.

Suggested criteria:

- Re-authorizing a user with thousands of grants for other clients does not scan/read all of those unrelated grants.
- The number of KV operations in one authorization request is bounded by a documented constant or by the number of grants for the same user+client, not total grants for the user.
- Existing stale grants for the same user+client are still revoked reliably, or any delayed/best-effort cleanup semantics are explicitly documented and tested.
- Tests cover high-cardinality users with many unrelated client grants.
- Tests cover multiple old grants for the same user+client.
- Tests cover missing/expired grant records and stale index entries if an index is added.
- README/API docs explain any new cleanup/index behavior that users need to know about.
- Storage schema documentation is updated if new KV keys are introduced.

## Related work

- Follow-up from #231

mattzcarey:
## Summary

Pluggable storage for the provider, rebuilt on top of the JWT access token stack and cut down to what the provider actually needs. Two built-in adapters:

- **Workers KV**: the default when `storage` is omitted. Preserves today's physical keys and JSON values exactly, so existing deployments need no migration.
- **Durable Object SQLite**: one object per user for grants, tokens, and grant transitions, one per registered client, sharded replay markers. Strong read-after-write and fenced code/refresh transitions, which removes the KV read-after-write race reported on #237. Brand new, so it has a single current schema with no migrations, versions, or legacy decoding.

## Changes

- `src/storage/index.ts`: a small contract (`clients`, `grants`, `accessTokens`, `replay`, `maintenance` stores, grant transitions, `OAuthStorageError`). Records cross the boundary in their canonical `ClientInfo`/`Grant`/`Token` shapes; stores never hide expired records, the provider keeps its existing expiry checks.
- Every persistence call in `src/oauth-provider.ts` goes through `storage.open(env)`; no direct `OAUTH_KV` access remains in `src`. Opening is cheap and stateless, so there is no request-scoped connection, bound environment, or lifecycle.
- Code and refresh exchange use begin/commit transitions. They begin after validation, the `tokenExchangeCallback`, and the access-token format policy, so a failure there leaves the credential retryable. Under Durable Objects a second concurrent exchange of the same credential gets `busy`.
- Closes #232: optional `grants.listByUserAndClient`. The Durable Object adapter serves it from a partial index (a test asserts the query plan via `runInDurableObject`); KV keeps the user-prefix scan and filter.
- Retryable storage failures at the token endpoint return `temporarily_unavailable` with `Retry-After`. Token and grant TTL minimums come from the adapter (`60` for KV, `0` for the Durable Object).
- Additive public API: `storage` option, re-exported `OAuthStorageError`, subpath exports `./storage`, `./storage/kv`, `./storage/durable-object`.
- The Durable Object adapter is tested with `@cloudflare/vitest-pool-workers` against a real SQLite-backed Durable Object in workerd (`npm run test:workers`, also run by `npm test`). Adding that dependency bumps `wrangler` and `@cloudflare/workers-types`; the newer `ExecutionContext` type gained `abort()`, hence the two-line mock changes in existing tests.

## Design notes for review

- The Durable Object adapter has no global index, so `listClients`, `deleteClient`, and `purgeExpiredData` reject with `unsupported_operation` on that adapter. Documented in `docs/storage-providers.md`.
- The Durable Object provider ships as **experimental** (9d1009b): the changeset, README, `docs/storage-providers.md`, and the factory's JSDoc say it is new with no production mileage and that its object schema may change in a minor release. Workers KV is the stable default.
- Refresh token reuse detection (#214) is not implemented. The grace window policy is unchanged.
- `open(env)` is called per operation, as the contract says, with no per-environment cache: a provider that resolves bindings per request always sees the current environment (8a7419e).
- Workers KV cannot fence overlapping exchanges of the same refresh token: it has no compare-and-swap and its reads are eventually consistent, so that case stays last-write-wins as on `main`. The Durable Object adapter takes a fenced lease per grant and answers the second exchange with `temporarily_unavailable` and `Retry-After`.
- Rebased onto the current #292 (after its review round). Behaviour carried over from #289 and #292 into the storage layer: the enterprise-managed grant builds its token before the `jti` replay marker is reserved, so a signing failure leaves the assertion retryable; the code grant builds (and for JWTs, signs) the access token inside the transition before commit, so a signing failure leaves the code retryable; a subject token too close to expiry to exchange stays `invalid_request` (RFC 8693 §2.2.2) with the threshold now read from the adapter's minimum TTL; the KV adapter's `deleteWithGrants` deletes the client record first and reads each key listing to completion before deleting anything.

## Stack

Stacked on #292, which is stacked on #289. Targets `feat/self-contained-access-tokens`. The examples PR (#314) sits on top of this one.

## Verification

```
npm run check            Node suite: Test Files 14 passed, Tests 997 passed (unit + workerd conformance)
                         Durable Object suite (workerd): Test Files 1 passed, Tests 7 passed
npm run build            complete; emits dist/oauth-provider.js and dist/storage/{index,kv/index,durable-object/index}.js
npx prettier --check .   clean
```


mattzcarey:
On the **Concurrent refresh discards issued token** finding: this is the Workers KV consistency model, not a change in this PR. KV has no compare-and-swap and reads are eventually consistent, so two exchanges of the same refresh token that overlap are last-write-wins there, exactly as on `main` today. The KV adapter's transition is a re-check, not a lock, and it deliberately honours the previous refresh token so a client that lost the response can retry (the grace window policy). A snapshot fence on the KV commit would still read possibly stale data, so it would only narrow the window while adding a new failure mode. The Durable Object adapter is the fix: it takes a fenced lease per grant, and the second overlapping exchange gets `temporarily_unavailable` with `Retry-After`. Adding this to the design notes in the description.

mattzcarey:
Follow-up in 8a7419e on the **Cached stores bypass fresh resolution** finding: `openStorage()` no longer caches per environment object. The contract already says `open(env)` must be cheap and is called per operation, so a provider that resolves bindings per request now always sees the current environment. A test wraps the KV provider in a counting one, sends two requests with the same env object, and checks `open` ran for both.

mattzcarey:
Closing: we're not building storage on Durable Objects. The storage provider we want is **Postgres via Hyperdrive**, which also gives real transactions where KV only has best-effort ones: atomic single use for consent and upstream transactions (the non-atomic read-then-delete documented in 1.1.0's `oauth-consent.ts`), and serialized refresh-token rotation (#214). That will be designed fresh against 1.x's storage (grant key metadata from #317, `transaction:` records from #335) rather than rebased from here.

